PR #1, opened 2016-02-16
Part 1: Added navigation rules and subtask steps for general case.

syoung-smallwisdom:
The intention with this work is to come up with a generalized architecture for setting up navigation rules and creating steps. This will be applied first in onboarding to setup navigation for eligibility and the consent quiz, but it can also be extended to the SBBSurvey classes used by the BridgeSDK.


Erin-Mounts:
looks good

PR #5, opened 2016-03-03
Add hook to instruction steps to allow these to point to a learn more with HTML content

syoung-smallwisdom:
…


Erin-Mounts:
looks good

PR #12, opened 2016-03-29
BRIDGE-1226 Added logic for user registration and login

syoung-smallwisdom:
I moved the demo code files into the sample app, but I have not updated the sample app (yet) to use the implementation defined in the SDK.

Additionally, I implemented signup, signin and consent for the user.


BridgeAppSDK/Localization.swift
```diff
@@ -49,15 +53,81 @@ class Localization: NSObject {
...
+        }
+        
+        let result =
+            trim(bundleInfo? ["CFBundleDisplayName"]) ??
+            trim(bundleInfo? ["CFBundleName"]) ??
+            trim(bundleInfo? ["CFBundleExecutable"]) ??
+            trim(mainBundle.objectForInfoDictionaryKey("CFBundleDisplayName")) ??
+            trim(mainBundle.objectForInfoDictionaryKey("CFBundleName")) ??
+            trim(mainBundle.objectForInfoDictionaryKey("CFBundleExecutable")) ??
+            trim(localizedBundleInfo? ["CFBundleDisplayName"]) ??
```
Erin-Mounts: shouldn't the app name from localizedInfoDictionary take precedence if it exists? (or am I misunderstanding the syntax?)

BridgeAppSDK/SBALocalizationMacroWrapper.m
```diff
@@ -0,0 +1,47 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+#import "SBALocalizationMacroWrapper.h"
+@import ResearchKit;
+
+@implementation SBALocalizationMacroWrapper
+
++ (NSString *)localizedORKString:(NSString *)orkStringKey {
+    return ORKLocalizedString(orkStringKey, nil);
```
Erin-Mounts, line 40: Does BridgeAppSDK directly access ResearchKit strings? Or has the method/function the macro expansion uses for finding the bundle to use for localization been implemented or overridden in BridgeAppSDK to point to our bundle? Or...?

BridgeAppSDK/SBANavigableOrderedTask.swift
```diff
@@ -116,11 +116,11 @@ public class SBANavigableOrderedTask: ORKOrderedTask {
         // Look for step in the ordered steps and lop off everything after this one
         if let previousIdentifier = step?.identifier,
             let idx = self.orderedStepIdentifiers.indexOf(previousIdentifier) where idx < self.orderedStepIdentifiers.endIndex {
-                self.orderedStepIdentifiers.removeRange(Range(start: idx.advancedBy(1), end: self.orderedStepIdentifiers.endIndex))
+                self.orderedStepIdentifiers.removeRange(idx.advancedBy(1) ..< self.orderedStepIdentifiers.endIndex)
```
Erin-Mounts, line 119: wouldn't this mean remove from idx + 1 to endIndex - 1 (..<)? Or does that operator mean "up to and including"?

BridgeAppSDK/SBAResourceFinder.swift
```diff
@@ -89,6 +91,14 @@ class SBAResourceFinder: NSObject {
         return nil
     }
     
+    func plistNamed(resouceNamed: String) -> NSDictionary? {
```
Erin-Mounts: missing the 2nd 'r' in 'resourceNamed'

BridgeAppSDK/SBASubtaskStep.swift
```diff
@@ -55,7 +55,7 @@ public class SBASubtaskStep: ORKStep {
     
     func substepIdentifier(identifier: String) -> String? {
         guard let range = identifier.rangeOfString("\(self.subtask.identifier).") else { return nil }
-        let stepRange = Range(start: range.endIndex, end: identifier.endIndex)
+        let stepRange = range.endIndex ..< identifier.endIndex
```
Erin-Mounts, line 58: I'm guessing I've got the meaning of the ..< operator wrong, then...

BridgeAppSDK/SBAUser.swift
```diff
@@ -0,0 +1,308 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import UIKit
+import BridgeSDK
+import ResearchKit
+
+public class SBAUser: NSObject, SBAUserWrapper {
+    
+    let lockQueue = dispatch_queue_create("com.sba.UserLockQueue", nil)
```
Erin-Mounts: should use org.sagebase to start the queue name to ensure no conflicts with queues 3rd parties might create if they build apps based on this--maybe org.sagebase.sba to ensure no conflicts with queues we create in other modules

BridgeAppSDK/SBAUser.swift
```diff
@@ -0,0 +1,308 @@
...
+
+    public func logout() {
+        self.sessionToken = nil
+        dispatch_sync(lockQueue) {
```
Erin-Mounts: If the things being set in the lockQueue can also only be accessed via the lockQueue, better to use dispatch_async except when you're going to return a value you're pulling from the lock queue (e.g. in getKeychainObject()), to minimize blocking (and reduce possibility of inadvertent deadlock as call chains evolve and become more complex, methods get overridden, etc.).

BridgeAppSDK/SBAUser.swift
```diff
@@ -0,0 +1,308 @@
...
+            obj = ORKKeychainWrapper.objectForKey(key, error: &err)
+            if let error = err {
+                print("Error accessing keychain: \(error)")
+            }
+        }
+        return obj
+    }
+    
+    func setKeychainObject(object: NSSecureCoding?, key: String) {
+        dispatch_sync(lockQueue) {
```
Erin-Mounts: dispatch_async

BridgeAppSDK/SBAUser.swift
```diff
@@ -0,0 +1,308 @@
...
+    func syncBoolForKey(key: String) -> Bool {
+        var ret: Bool = false
+        dispatch_sync(lockQueue) {
+            ret = self.userDefaults().boolForKey(key)
+        }
+        return ret
+    }
+    
+    func syncSetBool(value:Bool, forKey key: String) {
+        dispatch_sync(lockQueue) {
```
Erin-Mounts: dispatch_async

BridgeAppSDK/SBAUser.swift
```diff
@@ -0,0 +1,308 @@
...
+    func syncIntForKey(key: String) -> Int {
+        var ret: Int = 0
+        dispatch_sync(lockQueue) {
+            ret = self.userDefaults().integerForKey(key)
+        }
+        return ret
+    }
+    
+    func syncSetInteger(value:Int, forKey key: String) {
+        dispatch_sync(lockQueue) {
```
Erin-Mounts: dispatch_async

BridgeAppSDK/SBAUser.swift
```diff
@@ -0,0 +1,308 @@
...
+    func syncObjectForKey(key: String) -> AnyObject? {
+        var ret: AnyObject?
+        dispatch_sync(lockQueue) {
+            ret = self.userDefaults().objectForKey(key)
+        }
+        return ret
+    }
+    
+    func syncSetObject(value:AnyObject?, forKey key: String) {
+        dispatch_sync(lockQueue) {
```
Erin-Mounts: dispatch_async

BridgeAppSDK/SBAUserBridgeManager.m
```diff
@@ -0,0 +1,176 @@
...
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+#import "SBAUserBridgeManager.h"
+
+@implementation SBAUserBridgeManager
+
++ (BOOL) serverDisabled {
+#if DEVELOPMENT
```
Erin-Mounts: I believe this whole serverDisabled thing is a fossil from Y Media Labs related to mocking out the entire Bridge backend during development that isn't used anymore, and likely doesn't still work assuming it ever did (and was a wrongheaded approach to their development in the first place imo, led to a lot of churn when trying to integrate with our SDK later). That said, this may or may not be an opportune time to give it a decent burial (or burn it in all the fires, which would be my preference).

BridgeAppSDK/SBAUserWrapper+Bridge.swift
```diff
@@ -0,0 +1,311 @@
...
+     * Register a new user with an email/password
+     */
+    public func registerUser(email email: String, password: String, externalId: String?, dataGroups dataGroupsIn: [String]?, completion: ((NSError?) -> Void)?) {
+        
+        func completeRegistration(isTester: Bool) {
+            
+            // include test_user in the data groups if applicable
+            var dataGroups: [String]? = dataGroupsIn
+            if (isTester) {
+                dataGroups = (dataGroups ?? []) + [SBATestDataGroup]
```
Erin-Mounts, line 85: Are we getting SBATestDataGroup from the same plist where we get the study identifier etc.? If not, maybe we should, so it can be customized per study.

BridgeAppSDK/SBAUserWrapper+Bridge.swift
```diff
@@ -0,0 +1,311 @@
...
+            let error = NSError(domain: SBAUserErrorDomain, code: code, userInfo: nil)
+            self.callCompletionOnMain(error, completion: completion)
+            return
+        }
+        
+        // Make sure that 
+        SBAUserBridgeManager.ensureSignedInWithCompletion { [weak self] (_, error) in
+            guard (self != nil) else { return }
+            
+            if (error != nil) && error!.code == SBBErrorCode.ServerPreconditionNotMet.rawValue {
```
Erin-Mounts, line 188: FYI if you're already signed in (i.e. have a session token), no attempt will be made to sign you in again and error will be nil here. You'll still get the 412 on the next attempt to call an API that requires consent, and it'll be handled then, but as I said, just FYI in case it wasn't obvious.

BridgeAppSDK/SBAUserWrapper+Bridge.swift
```diff
@@ -0,0 +1,311 @@
...
+            else {
+                // otherwise, we are done. Set the flag that the consent has been verified and 
+                // call the completion
+                self!.consentVerified = !requiresConsent
+                self!.callCompletionOnMain(error, completion: completion)
+            }
+        }
+    }
+    
+    private func copyFromUserSession(response: SBAUserSessionInfoWrapper) {
```
Erin-Mounts: probably I'm just conditioned from my pre-ARC days using Core Foundation, but I feel like this should be named "updateFrom..." or some such rather than "copyFrom...".

BridgeAppSDK/SBAUserWrapper+Bridge.swift
```diff
@@ -0,0 +1,311 @@
...
+    var dataGroups : [String]? {
+        return self["dataGroups"] as? [String]
+    }
+    
+    var dataSharingEnabled : Bool {
+        return self["dataSharing"] as? Bool ?? false
+    }
+    
+    var dataSharingScope: SBBUserDataSharingScope {
+        guard let sharingKey = self["dataGroups"] as? String where self.dataSharingEnabled else {
```
Erin-Mounts: I'm confused about the syntax here since I thought self["dataGroups"] should be an array of strings, not a single string... can you enlighten me as to what's going on here?

BridgeAppSDK/SBAUserWrapper+Bridge.swift
```diff
@@ -0,0 +1,311 @@
...
+            return .None
+        }
+        return SBBUserDataSharingScope(key: sharingKey)
+    }
+
+    var subpopulationGuid: String? {
+        // TODO: Handle multiple consent groups with separate sub populations
+        if let consentStatuses = self["consentStatuses"] as? [String : [String : AnyObject]] {
+            for (_, subpop) in consentStatuses {
```
Erin-Mounts, line 297: remind me what the '_' means here?

BridgeAppSDK/SBAUserWrapper.swift
```diff
@@ -0,0 +1,113 @@
...
+//
+
+import Foundation
+
+// Wrapper support for reverse compatiblility to AppCore. syoung 03/28/2016
+
+public protocol SBAUserWrapper: class, SBBAuthManagerDelegateProtocol {
+
+    /**
+     * SessionToken is stored in memory only
```
Erin-Mounts, line 41: Didn't it used to be kept in the keychain too? Any particular reason not to now (or am I mistaken about that)?

BridgeAppSDK/Localization.swift
syoung-smallwisdom: Good point. This was a copy/paste from AppCore that I included because it seemed to be that not all apps were defining the app name in the expected place. I tilled my head and then got distracted and forgot about it. I'll rearrange.

BridgeAppSDKSample/HKHealthStore+Queries.swift
```diff
@@ -32,7 +32,7 @@ import HealthKit
 
 extension HKHealthStore {
     /// Asynchronously fetches the most recent quantity sample of a specified type.
-    func mostRecentQauntitySampleOfType(quantityType: HKQuantityType, predicate: NSPredicate? = nil, completion: (HKQuantity?, NSError?) -> Void) {
+    func mostRecentQuantitySampleOfType(quantityType: HKQuantityType, predicate: NSPredicate? = nil, completion: (HKQuantity?, NSError?) -> Void) {
```
Erin-Mounts, line 35: thank you

BridgeAppSDK/SBALocalizationMacroWrapper.m
syoung-smallwisdom, line 40: swift cannot access macros so anything that uses a macro needs to be wrapped. In this case, ORKLocalizedString looks in the researchKit bundle for the localized strings. I am using it because  there are certain cases where we want consistency in wording (Ok, Cancel, Yes, No, etc).

BridgeAppSDKSample/OnboardingViewController.swift
```diff
@@ -13,6 +13,7 @@ other materials provided with the distribution.
 
 3.  Neither the name of the copyright holder(s) nor the names of any contributors
 may be used to endorse or promote products derived from this software without
+ @IBOutlet weak var registrationCodeTextField: UITextField!
```
Erin-Mounts: inadvertent insert?

BridgeAppSDK/SBALocalizationMacroWrapper.m
Erin-Mounts, line 40: gotcha, makes sense.

BridgeAppSDK/SBASubtaskStep.swift
syoung-smallwisdom, line 58: Good question. Adding a test. I followed the instructions in the deprecation warning (and google) and unit tests still pass, but it doesn't really make sense. Checking now and will either update or just include the test.

Erin-Mounts, line 58: I suppose endIndex might point to just beyond the end of the range, like using the length of a string or array.count as an index...

Erin-Mounts, line 58: actually seeing it in this context, that would fit--start just past the end of the matched substring, and go to the end of the whole identifier string by stopping just before its end marker.

BridgeAppSDK/SBAUserWrapper+Bridge.swift
syoung-smallwisdom, line 188: I am relying upon the assumption that this is the call that gets made when a user launches their app after an update (that might include requiring re-consent). Unfortunately, there isn't a clean way to handle all possible scenarios for all apps b/c we do _not_ want to clear the stored consentSignature if this is a user who is in a state where they are either (a) validating a new registration or (b) logging in with an existing account (that may need a new consent).

BridgeAppSDK/SBAUserWrapper+Bridge.swift
syoung-smallwisdom: I like your method name better. Changing.

BridgeAppSDK/SBAUserWrapper+Bridge.swift
syoung-smallwisdom: copy/paste error. thanks for catching.

BridgeAppSDK/SBAUserWrapper+Bridge.swift
syoung-smallwisdom, line 297: unused

BridgeAppSDK/SBAUserWrapper.swift
syoung-smallwisdom, line 41: No. It _says_ it is stored in the keychain, but it actually was only stored in memory.

BridgeAppSDKSample/OnboardingViewController.swift
syoung-smallwisdom: maybe?

BridgeAppSDK/SBASubtaskStep.swift
syoung-smallwisdom, line 58: Double checked to make sure that unit tests really covered this and added a test to ensure that they both do the same thing.

BridgeAppSDK/SBAUserWrapper+Bridge.swift
syoung-smallwisdom, line 85: we are now

syoung-smallwisdom:
@Erin-Mounts 
I think I've got everything covered.

BridgeAppSDKSample/OnboardingViewController.swift
Erin-Mounts: ...?

Erin-Mounts:
looks good except for the extraneous line in the license comment block at the top of BridgeAppSDKSample/OnboardingViewController.swift:

`+ @IBOutlet weak var registrationCodeTextField: UITextField!`

PR #13, opened 2016-04-09
Bridge 1230

syoung-smallwisdom:
Requires: https://github.com/Sage-Bionetworks/ResearchKit/pull/23

Migration of Medication Tracking and allowing for creation of a task using the ORKOrderedTask factory.


BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -93,27 +92,39 @@ public class SBABridgeInfoPList : NSObject, SBABridgeInfo {
...
+        }
+        return email.stringByReplacingCharactersInRange(range, withString: "+%@@")
+    }
     
+    public var appStoreLinkURL: NSURL! {
+        guard let appStoreLinkURLString = self.appStoreLinkURLString,
+            let url = NSURL(string: appStoreLinkURLString)
+            else {
+                return NSBundle.mainBundle().appStoreLinkURL()
+        }
```
Erin-Mounts, line 125: this curly brace formatting looks wrong even though it's syntactically correct. not sure there's a cleaner way but thought I'd mention it since it "threw me out of the story" so to speak...

BridgeAppSDK/SBABridgeTask+Dictionary.swift
```diff
@@ -0,0 +1,81 @@
...
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import Foundation
+
+extension NSDictionary: SBABridgeTask {
+    
+    public var taskIdentifier: String! {
+        return _nullableTaskIdentifier ?? NSUUID().UUIDString
```
Erin-Mounts: if there's no @"taskIdentifier" key in the underlying dictionary, won't this give you a different value each time you access the taskIdentifier for this SBABridgeTask instance?

BridgeAppSDK/SBABridgeTask.swift
```diff
@@ -0,0 +1,90 @@
...
+        }
+        
+        // Map the insert steps
+        if let insertSteps = self.insertSteps?.enumerate().map({ (index, item) in
+            return item.transformToStep(factory, isLastStep: false)
+        }) where insertSteps.count > 0 {
+            
+            var introStep: ORKStep!
+            if let subtaskStep = subtaskSteps.first as? SBASubtaskStep,
+                let orderedTask = subtaskStep.subtask as? ORKOrderedTask {
```
Erin-Mounts, line 66: Will this work with nested subtasks arbitrarily deep as the first step? or is there a limit on allowed subtask depth?

BridgeAppSDK/SBABridgeInfo.swift
syoung-smallwisdom, line 125: It's definitely one of the things I am struggling with in swift especially since the lines can get so long. I've taken to putting the `else` on the next line if I have multiple `let` or `where` statements, and to using the put the curly black in the same place everywhere since that's what Apple does in their examples.

BridgeAppSDK/SBABridgeTask+Dictionary.swift
syoung-smallwisdom: whoops, yes.

BridgeAppSDK/SBADataObject.m
```diff
@@ -0,0 +1,187 @@
...
+@interface SBADataObject ()
+
+@property (nonatomic, copy, readwrite) NSString * classType;
+
+@end
+
+@implementation SBADataObject
+
+- (instancetype)init {
+    [NSException exceptionWithName: NSInternalInconsistencyException
```
Erin-Mounts: exception created but never actually @thrown -- or did you mean to "raise" it?

BridgeAppSDK/SBAExternalIDOnboardingController.swift
```diff
@@ -0,0 +1,82 @@
...
+public extension SBAExternalIDOnboardingController {
+    
+    public func handleViewDidAppear() {
+        registrationCodeTextField.becomeFirstResponder()
+    }
+
+    public func registerUser() {
+        
+        guard let text: String = self.registrationCodeTextField.text
+            where text.lengthOfBytesUsingEncoding(NSUTF8StringEncoding) > 0 else {
```
Erin-Mounts: why not just 'where text.length > 0' ?

BridgeAppSDK/SBABridgeTask.swift
syoung-smallwisdom, line 66: No, not intended to be recursive. I could make it recursive (using a block) but that also makes some already obfuscated code even more difficult to grok.

BridgeAppSDK/SBAExternalIDOnboardingController.swift
syoung-smallwisdom: not an available method in swift.

BridgeAppSDK/SBAJSONObject.m
```diff
@@ -0,0 +1,257 @@
...
+@end
+
+@implementation NSDate (SBAJSONObject)
+
+- (id)jsonObjectWithFormatter:(NSFormatter * _Nullable)formatter {
+    NSDateFormatter *dateFormatter = [formatter isKindOfClass:[NSDateFormatter class]] ? (NSDateFormatter *)formatter : [self defaultFormatter];
+    return [dateFormatter stringFromDate:self];
+}
+
+- (NSDateFormatter *)defaultFormatter {
```
Erin-Mounts: BridgeSDK has a public category (NSDate+SBBAdditions.h) to handle Bridge-compliant conversions between NSDate objects and ISO8601 date strings. Possible to splice in here instead of duplicating that functionality?

syoung-smallwisdom: Yes, this was from AppCore. And still somewhat stubbed out code. (Started to work on it but ended up distracted by Med tracking once I found out we needed that.)

BridgeAppSDK/SBASurveyItem+Dictionary.swift
```diff
@@ -33,15 +33,36 @@
...
+            return factory.createSurveyStep(self, isSubtaskStep: nil, isLastStep: isLastStep)
+        }
+    }
+}
+
 extension NSDictionary: SBASurveyItem {
     
     public var identifier: String! {
-        return (self["identifier"] as? String) ?? "\(self.hash)"
+        return (self["identifier"] as? String) ?? (self["schemaIdentifier"] as? String) ?? NSUUID().UUIDString
```
Erin-Mounts: different value every time you access this instance's identifier if no identifier or schemaIdentifier keys in the underlying dictionary

BridgeAppSDK/SBATrackedDataObjectCollection.m
```diff
@@ -0,0 +1,110 @@
...
+- (NSNumber *)schemaRevision {
+    if (_schemaRevision == nil) {
+        _schemaRevision = @(1);
+    }
+    return _schemaRevision;
+}
+
+- (SBATrackedDataStore *)dataStore {
+    if (_dataStore == nil) {
+        return [SBATrackedDataStore defaultStore];
```
Erin-Mounts, line 64: you don't want to set the property to the default store?

syoung-smallwisdom, line 64: I think this was a muscle memory thing to make sure it didn't end up in a retain loop? Not sure.

BridgeAppSDK/SBABridgeTask.swift
Erin-Mounts, line 66: later, if actually needed, then

BridgeAppSDK/SBATrackedDataObjectCollection.m
Erin-Mounts, line 64: ah ok didn't notice it was a thing that could end up pointing to itself... makes sense now

syoung-smallwisdom, line 64: That's because it doesn't, but in one iteration of the implementation that I have in my head, it did. The data collection needs the data store to figure out which steps to include. I haven't gotten there yet and I'm not sure what I will do. In the Parkinson's app it uses a bunch of hardcoded keys which are accessible from the dashboard, thus the data store doesn't need a pointer to the data collection (which is _not_ retained outside of being used to run a task). Here, the ids that the data store (and dashboard) need are stored in the json file (used to create the data collection). Thus, a possible loop. But I haven't actually updated mPowerSDK to not use hardcoded keys so currently there is no loop there, either.

syoung-smallwisdom, line 64: I'm actually working on that bit now, but it's not ready. And the property is now being set.

Erin-Mounts:
Looks good

PR #14, opened 2016-04-14
More updates to allow medication tracking and task creation

syoung-smallwisdom:


BridgeAppSDK/SBAResourceFinder.swift
```diff
@@ -93,9 +93,14 @@ class SBAResourceFinder: NSObject {
...
     func jsonNamed(resourceNamed: String, bundle: NSBundle?) -> NSDictionary? {
-        if let data = self.dataNamed(resourceNamed, ofType: "json", bundle: bundle),
-            let json = try? NSJSONSerialization.JSONObjectWithData(data, options: .MutableContainers){
+        do {
+            if let data = self.dataNamed(resourceNamed, ofType: "json", bundle: bundle) {
+                let json = try NSJSONSerialization.JSONObjectWithData(data, options: .MutableContainers)
                 return json as? NSDictionary
+            }
+        }
+        catch let error as NSError {
```
Erin-Mounts, line 102: Since this is getting JSON resources from the bundle, a read or parse error would indicate a programming error. If you're going to catch it rather than letting it crash, I'd prefer it to do something more noticeable than putting an inconspicuous message in the (fairly noisy in debug builds) console log.

Erin-Mounts:
one nitpick about using catch for what would be a programming error, otherwise looks good

BridgeAppSDK/SBAResourceFinder.swift
syoung-smallwisdom, line 102: Good point. Fixed.

PR #15, opened 2016-04-14
Moved stubbed out implementation of the activities view controller into BridgeApp

syoung-smallwisdom:
I went ahead and moved the Activities view controller into BridgeAppSDK. This is not a complete implementation (does not include "keep going" section or handling of SBBSurvey objects or archiving and saving results for display in dashboard. (etc.) But it should give you an idea where I'm thinking the scheduling stuff could go and makes it easier us to implement concurrently.


Erin-Mounts:
Looks good

PR #16, opened 2016-04-15
Added permissions and cleaned up some of the activities

syoung-smallwisdom:
Note: there is some stubbed out code and some less than ideal copy/paste from the AppCore implementation. Would like to see a different implementation of permission handling for HealthKit (ironically) b/c I think that the AppCore implementation is unnecessarily obfuscated, but that's a project for another day.

Adding to open pull request. Now dependent on https://github.com/Sage-Bionetworks/ResearchKit/pull/25

### DIFF
--- a/BridgeAppSDK/SBATaskViewController.m
+++ b/BridgeAppSDK/SBATaskViewController.m
@@ -1,0 +1,72 @@
+//
+//  SBATaskViewController.m
+//  BridgeAppSDK
+//
+//  Copyright © 2016 Sage Bionetworks. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
+//
+// 1.  Redistributions of source code must retain the above copyright notice, this
+// list of conditions and the following disclaimer.
+//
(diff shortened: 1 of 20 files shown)


BridgeAppSDK/SBATaskViewController.m
```diff
@@ -0,0 +1,72 @@
...
+
+#import "SBATaskViewController.h"
+#import <BridgeAppSDK/BridgeAppSDK-Swift.h>
+
+@implementation SBATaskViewController
+
+- (NSURL *)outputDirectory {
+    NSURL *outputDirectory = [super outputDirectory];
+    if (outputDirectory == nil) {
+        NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
```
Erin-Mounts: I'm not sure it matters since we don't enable iTunes file sharing for our apps, but the application support directory might be a more appropriate place for task result files since they're not meant to be user-accessible. (I'm assuming that's what this is for...)

Erin-Mounts:
just the one maybe-relevant comment about where to put output files, otherwise looks good

Erin-Mounts:
looks good

PR #18, opened 2016-04-26
Make Localization more subclassable

Erin-Mounts:
to support localization at the app level without recreating everything


syoung-smallwisdom:
looks good

PR #20, opened 2016-04-28
refactor

Erin-Mounts:
sign in actually uses email--username is deprecated and must be equal to the email field if used--so let's refactor to reflect that. Requires https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/64


syoung-smallwisdom:
looks good

PR #29, opened 2016-05-03
BRIDGE-1301 Allow conditional hiding of the cancel button

syoung-smallwisdom:


BridgeAppSDK/ORKOrderedTask+SBAExtension.h
```diff
@@ -0,0 +1,45 @@
...
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+#import <ResearchKit/ResearchKit.h>
+
+@protocol SBATaskExtension <ORKTask>
+
+- (NSUInteger)stepCount;
+- (NSUInteger)indexOfStep:(ORKStep *)step;
```
Erin-Mounts, line 39: is this still needed here since you made it public in the "base" class? I notice it's not in the category's implementation...

syoung-smallwisdom, line 39: This is a protocol extension of `ORKTask` which is not to be confused with `ORKOrderedTask`. ;)

Basically, I didn't want to force implementations of ORKTask that are valid for a task view controller to _have_ to inherit from ORKOrderedTask (even though most of the tasks we use do so).

BridgeAppSDK/SBAActivityTableViewController.swift
```diff
@@ -156,6 +167,28 @@ public class SBAActivityTableViewController: UITableViewController, SBASharedInf
     
     // MARK: Protected subclass methods
     
+    public func shouldShowTaskForSchedule(schedule: SBBScheduledActivity) -> Bool {
+        return true
```
Erin-Mounts, line 171: Is this a stub to be filled in later, or a base class implementation to be overridden?

syoung-smallwisdom, line 171: Need to be able to conditionally override to allow Lilly to disallow doing the activity _again_ for some cases but _not_ others.

Update: Will edit in next PR to disallow showing task for expired.

Erin-Mounts:
looks good other than those two comments, merge when ready

PR #30, opened 2016-05-03
BRIDGE-1307 Check the title override for a spoken instruction that matches.

syoung-smallwisdom:
This ensures that the spokenInstruction matches the title. Not sure if there is other work implied by the lira ticket so left it open and assigned to @Erin-Mounts

### DIFF
--- a/BridgeAppSDK/SBAActiveTask.swift
+++ b/BridgeAppSDK/SBAActiveTask.swift
@@ -119,17 +119,16 @@ extension SBAActiveTask {
         if let items = self.localizedSteps {
             for item in items {
                 if let step = task.steps.filter({ return $0.identifier == item.identifier }).first {
-                    if let stepTitle = item.stepTitle {
-                        if let activeStep = step as? ORKActiveStep where activeStep.spokenInstruction == step.title {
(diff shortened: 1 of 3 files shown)


Erin-Mounts:
The spoken text for the gait/balance task was correct, the written text should match it by mentioning the turning in a circle. I don't know that we want to enforce the written and spoken text _always_ being the same, but Lilly wanted it here because they found it confusing that the instructions seemed different (and I think largely because of the audio issue so they couldn't hear the spoken instructions during the task). As for the part about checking with Andrew, that was to verify that the participant is supposed to turn all the way around ("in a circle") and face the same way, not just "turn around" (180º), i.e. that the current spoken instructions are accurate.

syoung-smallwisdom:
Ok. _sigh_

They had previously specifically asked that the "in a circle" be taken out. and I did so for the written instruction (via change to CombinedTask.json file) but forgot about the spoken instruction.

syoung-smallwisdom:
@Erin-Mounts 
Changed to instead allow the spokenInstruction to also be changed independently of the title.

Erin-Mounts:
looks good

PR #32, opened 2016-05-04
UI updates and fixes to tracked data rule handling

syoung-smallwisdom:


BridgeAppSDK/SBAActivityTableViewController.swift
```diff
@@ -163,8 +175,12 @@ public class SBAActivityTableViewController: UITableViewController, SBASharedInf
             where shouldRecordResult(schedule, taskViewController: taskViewController) {
             
             updateScheduledActivity(schedule, taskViewController: taskViewController)
+            taskViewController.task?.updateTrackedDataStores(true)
```
Erin-Mounts: not obvious what the parameter does until you look at the function definition--could you include the parameter name in the call?

Erin-Mounts:
one minor nit, otherwise looks good

PR #36, opened 2016-05-06
BRIDGE-1302 Update schedule to allow custom filtering

syoung-smallwisdom:
Refactored the Activity view controller to use a data manager so that the data manager could be overridden to customize functionality, but the bulk of the logic is in a data manager that can be used for either a tableview or a collectionview.


Erin-Mounts:
looks good

PR #42, opened 2016-05-10
BRIDGE-1317 Only create a notification for scheduled activities when …

syoung-smallwisdom:
…configured to do so.


Erin-Mounts:
looks good

PR #43, opened 2016-05-11
BRIDGE-1279 Split the task results into subgroups by activity

syoung-smallwisdom:
This only includes tests for cases where there is only one task result or where there are multiple results in the same activity. This does not include testing (yet) for the mPower app or for surveys from the server.


Erin-Mounts:
looks good

PR #46, opened 2016-05-18
Bridge 1279--archiving, encrypting, uploading

Erin-Mounts:


BridgeAppSDK/DataArchiving/SBADataArchive.h
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ Task Archive Initializer
+ 
+ @param     reference           Reference for the archive used as a directory name in temp directory
+ @param     schemaRevision      Schema revision to use for validating archived results when uploaded to Bridge
+ 
+ @return    APCDataArchive      An instance of APCDataArchive
+ */
+- (id)initWithReference:(NSString *)reference schemaRevision:(NSNumber *)schemaRevision;
```
syoung-smallwisdom: Use the `setArchiveInfoObject:forKey:` method instead which can be used to add any custom info to the dictionary.

BridgeAppSDK/DataArchiving/SBADataArchive.m
```diff
@@ -0,0 +1,280 @@
...
+        [self.infoDict setObject:[[NSBundle mainBundle] appName] forKey:kAppNameKey];
+        [self.infoDict setObject:[self.class appVersion] forKey:kAppVersionKey];
+        [self.infoDict setObject:[[UIDevice currentDevice] deviceInfo] forKey:kPhoneInfoKey];
+
+        // TODO: syoung 04/26/2016 Implement using setArchiveInfoObject: to add these values
+
+//        [self.infoDict setObject:[NSUUID new].UUIDString forKey:kTaskRunKey];
+        if (self.schemaRevision) {
+            [self.infoDict setObject:self.schemaRevision forKey:kSchemaRevisionKey];
+        }
```
syoung-smallwisdom: Use the setArchiveInfoObject:forKey: method instead which can be used to add any custom info to the dictionary.

BridgeAppSDK/DataArchiving/SBAEncryption.m
```diff
@@ -0,0 +1,214 @@
...
+        BIO_free(tbio);
+    
+    return returnValue;
+}
+
+#pragma mark - helpers
+
++ (NSString*) pemPath
+{
+    return [SBAEncryptionHelper pemPath];
```
syoung-smallwisdom: Remove commented out code or explain why commented out.

BridgeAppSDK/DataArchiving/SBAEncryptionHelper.swift
```diff
@@ -0,0 +1,69 @@
...
+import UIKit
+
+@objc public class SBAEncryptionHelper: NSObject {
+    
+    public class func pemPath() -> String? {
+        let sharedAppDelegate = UIApplication.sharedApplication().delegate as! SBASharedAppDelegate
+        let certificatePath = NSBundle.mainBundle().pathForResource(sharedAppDelegate.bridgeInfo.certificateName, ofType: "pem")
+        return certificatePath
+    }
+//
```
syoung-smallwisdom: Remove commented out code or add TODO with explanation for why it's there.

BridgeAppSDK/DataArchiving/SBAEncryptionHelper.swift
```diff
@@ -0,0 +1,69 @@
...
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import UIKit
+
+@objc public class SBAEncryptionHelper: NSObject {
+    
+    public class func pemPath() -> String? {
+        let sharedAppDelegate = UIApplication.sharedApplication().delegate as! SBASharedAppDelegate
```
syoung-smallwisdom:
```
guard let sharedAppDelegate = UIApplication.sharedApplication().delegate as? SBASharedAppDelegate else { return nil }
```

Otherwise, this will throw an exception in a unit test.

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+
+private var QuestionResultQuestionTextKey = "questionText"
+private var QuestionResultQuestionTypeKey = "questionType"
+private var QuestionResultUserInfoKey = "userInfo"
+
+enum ResultType: Int {
+    case URL, Dictionary, Data
+}
+
+class ArchiveableResult : NSObject {
```
syoung-smallwisdom, line 82: Would be cleaner to use a struct here with the result, resultType and filename defined as let. Then it is a clearly immutable object that is used to pass some bits of information.

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+
+extension ORKFileResult {
+    override func bridgeData(stepIdentifier: String) -> ArchiveableResult? {
+        guard let url = self.fileURL else {
+            return nil
+        }
+        var ext = url.pathExtension
+        if ext == nil || ext == "" {
+            ext = "json"
+        }
```
syoung-smallwisdom, line 149:
```
let filename = bridgifyFilename(self.identifier + "_" + stepIdentifier) + "." + ext!
```

Little easier to read.

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+        for value: NSValue in targetRectangles {
+            let rectangle = value.CGRectValue()
+            let stringified = NSStringFromCGRect(rectangle)
+            rectangles += [stringified]
+        }
+        return  rectangles
+    }
+}
+
+class AnswerKeyAndValue: NSObject {
```
syoung-smallwisdom: use a struct and lets

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+
+protocol ORKQuestionResultAnswerJSON {
+    func jsonSerializedAnswer() -> (AnswerKeyAndValue)?
+}
+
+private var _timeOnlyFormatter: NSDateFormatter? = nil
+
+extension ORKQuestionResult: ORKQuestionResultAnswerJSON {
+    
+    func timeOnlyFormatter() -> NSDateFormatter {
```
syoung-smallwisdom: There is a BridgeSDK formatter that has the time-only formatter that is expected for sending to server.

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+        ChoiceQuestionResult[kStartDateKey] = self.startDate
+        ChoiceQuestionResult[kEndDateKey] = self.endDate
+        ChoiceQuestionResult[QuestionResultUserInfoKey] = self.userInfo
+        if let answer = self.jsonSerializedAnswer() {
+            ChoiceQuestionResult[answer.key] = answer.value
+        }
+        return (ChoiceQuestionResult as NSDictionary).jsonObject()
+    }
+    
+    func jsonSerializedAnswer() -> (AnswerKeyAndValue)? {
```
syoung-smallwisdom: Why the parenthesis? was this originally a tuple?

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+        let dateTime = answer as? NSDate
+        switch self.questionType {
+        case .TimeOfDay:
+            return timeOnlyFormatter().stringFromDate(dateTime!)
+        case .DateAndTime:
+            return dateTime!.ISO8601String()
+        case .Date:
+            return dateTime!.ISO8601DateOnlyString()
+        default:
+            // TODO: Implement specific formatters for other questionTypes as needed
```
syoung-smallwisdom: sign and date TODO comments

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+    func jsonSerializedAnswer() -> (AnswerKeyAndValue)? {
+        let className = NSStringFromClass(self.classForCoder)
+        fatalError("jsonSerializedAnswer not implemented for \(className)")
+    }
+    
+    func jsonForAnswer(answer: AnyObject) -> AnyObject {
+        let dateTime = answer as? NSDate
+        switch self.questionType {
+        case .TimeOfDay:
+            return timeOnlyFormatter().stringFromDate(dateTime!)
```
syoung-smallwisdom: If I recall correctly, I think this questionType returns a NSDateComponents object.  Really, I think this is a bit overly complicated.  Currently, `ORKChoiceQuestionResult` is the only one defined and it has a question type of `ORKQuestionTypeSingleChoice` or `ORKQuestionTypeMultipleChoice`.  All you really need here is a convenience method for returning the default json for a given object:

```
    func jsonForAnswer(answer: AnyObject) -> AnyObject {
        if let jsonAnswer = answer as? SBAJSONObject {
            // If this object conforms to the protocol then return the default jsonObject 
            return jsonAnswer.jsonObject()
        }
        else if let object = answer as? NSObject {
            // If it inherits from NSObject, then return the description
            return object.description
        }
        // Otherwise return NSNull
        return NSNull()
    }
```

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+        }
+    }
+}
+
+extension ORKChoiceQuestionResult {
+    override func jsonSerializedAnswer() -> (AnswerKeyAndValue)? {
+        let choiceAnswers = self.choiceAnswers?.map { (answer) -> AnyObject in
+            return jsonForAnswer(answer)
+        }
+        return AnswerKeyAndValue.init(key: "choiceAnswers", value: choiceAnswers)
```
syoung-smallwisdom:
```
guard let choiceAnswers = self.choiceAnswers? else { return nil }
return AnswerKeyAndValue(key: "choiceAnswers", value: (choiceAnswers as NSArray).jsonObject())
```

Also, I think that surveys might be designed to look at "answer" for all values, rather than the property associated with the specific type. Should probably check that.

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -133,6 +143,14 @@ public class SBABridgeInfoPList : NSObject, SBABridgeInfo {
     public var schemaMap: [NSDictionary]? {
         return self.plist["schemaMapping"] as? [NSDictionary]
     }
+    
+    public var filenameMap: NSDictionary? {
```
syoung-smallwisdom: Is the filenameMap still used? If not, could we remove it?

BridgeAppSDK/SBAJSONObject.m
```diff
@@ -34,8 +34,24 @@
 #import "SBAJSONObject.h"
 @import BridgeSDK;
 
+@implementation NSObject (SBAJSONObject)
```
syoung-smallwisdom: Not really a huge fan of extending NSObject with an NSAssert. What is the intended use-case? Additionally, this could return `[self description]` rather than asserting or returning an empty dictionary.

BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -201,11 +201,15 @@ public class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORK
         return true
     }
     
+    public func isScheduleAvailableYet(schedule: SBBScheduledActivity) -> Bool {
```
syoung-smallwisdom: Can you remove the "yet"? `Tis a silly word.

BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -278,7 +282,18 @@ public class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORK
...
+                print("Archiving results failed")
+                return
+            }
+            for archive in archives {
+                archive.encryptAndUploadArchiveWithCompletion({ (error) in
+                    if error != nil {
+                        print("Error encrypting and uploading archive:\n\(error)")
+                    }
+                })
+            }
```
syoung-smallwisdom: Please put all this into archiveResults() as a void function. That way subclasses can override as needed and it is easier to refactor to allow customization that hasn't yet been thought of.

BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -382,13 +397,54 @@ public class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORK
     public func sendUpdatedScheduledActivities(scheduledActivities: [SBBScheduledActivity]) {
         SBAUserBridgeManager.updateScheduledActivities(scheduledActivities)
     }
-    
-    public func archiveResults(schedule: SBBScheduledActivity, taskViewController: ORKTaskViewController) {
+        
+    public func archiveResults(schedule: SBBScheduledActivity, taskViewController: ORKTaskViewController) -> [SBADataArchive]? {
```
syoung-smallwisdom: See comment above. Perhaps `buildResultArchives` ?

BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -382,13 +397,54 @@ public class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORK
...
         
         let results = activityResultsForSchedule(schedule, taskViewController: taskViewController)
         print(results)
         
-        // TODO: implement syoung 04/27/2016 Stubbed out accessor
+        var archives = [SBADataArchive]()
+        for activityResult in results {
+            if let activityResultResults = activityResult.results as? [ORKStepResult] {
+                let archive = SBADataArchive.init(reference: activityResult.schemaIdentifier, schemaRevision: activityResult.schemaRevision)
```
syoung-smallwisdom: remove the `init`. Also, see my comment above about using set method to add to userInfo dictionary. This will allow surveys to also add info to the userInfo without the SBADataArchive needing to care about those kinds of details.

BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -382,13 +397,54 @@ public class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORK
...
+                                archive.insertDictionaryIntoArchive(archiveableResult.result as! Dictionary, filename: archiveableResult.filename)
+                            case .Data:
+                                archive.insertDataIntoArchive(archiveableResult.result as! NSData, filename: archiveableResult.filename)
+                            }
+                        }
+                    }
+                }
+                
+                let error = archive.completeArchive()
+                if error != nil {
```
syoung-smallwisdom:
```
if let error = archive.completeArchive() {
```

BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -382,13 +397,54 @@ public class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORK
...
+                let archive = SBADataArchive.init(reference: activityResult.schemaIdentifier, schemaRevision: activityResult.schemaRevision)
+                for stepResult in activityResultResults {
+                    if let stepResultResults = stepResult.results {
+                        for result in stepResultResults {
+                            
+                            //Update date if needed
+                            if (result.startDate == nil) {
+                                result.startDate = stepResult.startDate;
+                                result.endDate = stepResult.endDate;
+                            }
```
syoung-smallwisdom: `result.startDate = result.startDate ?? stepResult.startDate ?? NSDate()`
It's not as pretty but what if stepResult.startDate is nil?

BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -382,13 +397,54 @@ public class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORK
...
+                            
+                            guard let archiveableResult = result.bridgeData(stepResult.identifier) else {
+                                assertionFailure("Something went wrong getting result to archive from result \(result.identifier) of step \(stepResult.identifier) of activity result \(activityResult.identifier)")
+                                archive.removeArchive()
+                                return nil
+                            }
+                            
+                            switch archiveableResult.resultType {
```
syoung-smallwisdom: This feels like it belongs on ArchiveResult instead. Is it here b/c you are considering grouping the results?

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+    func makeTargetRectangleRecords(targetRectangles: [NSValue]) -> NSArray
+    {
+        var rectangles = [String]()
+        
+        for value: NSValue in targetRectangles {
+            let rectangle = value.CGRectValue()
+            let stringified = NSStringFromCGRect(rectangle)
+            rectangles += [stringified]
+        }
+        return  rectangles
```
syoung-smallwisdom:
````
return targetRectangles.map({ NSStringFromCGRect($0.CGRectValue()) })
````

File: BridgeAppSDK/ORKResult+SBAExtension.swift [hunk 24]
```diff
@@ -0,0 +1,341 @@
...
+            return samples
+        }
+        for sample: ORKSpatialSpanMemoryGameTouchSample in touchSamples! {
+            
+            var aTouchSample = [String: AnyObject]()
+            
+            aTouchSample[kSpatialSpanMemoryTouchSampleTimeStampKey]   = sample.timestamp
+            aTouchSample[kSpatialSpanMemoryTouchSampleTargetIndexKey] = sample.targetIndex
+            aTouchSample[kSpatialSpanMemoryTouchSampleLocationKey]    = NSStringFromCGPoint(sample.location)
+            aTouchSample[kSpatialSpanMemoryTouchSampleIsCorrectKey]   = sample.correct
```
syoung-smallwisdom, line 251: use map here, too

BridgeAppSDK/ORKResult+SBAExtension.swift
syoung-smallwisdom, line 272: Use a struct.

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+        }
+        
+        return _timeOnlyFormatter!
+    }
+    override func bridgeData(stepIdentifier: String) -> ArchiveableResult? {
+        let filename = bridgifyFilename(self.identifier) + ".json"
+        return ArchiveableResult.init(result: self.jsonObject(), resultType: .Dictionary, filename: filename)
+    }
+    
+    public override func jsonObject() -> AnyObject {
```
syoung-smallwisdom: override? of what? does this build? A different method name might be more appropriate. `dictionaryRepresentation` ?

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+        var ChoiceQuestionResult = [String: AnyObject]()
```
syoung-smallwisdom: Please use camelCase

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,341 @@
...
+        ChoiceQuestionResult[QuestionResultQuestionTextKey] = self.questionText
+        ChoiceQuestionResult[QuestionResultQuestionTypeKey] = self.questionType.rawValue
+        ChoiceQuestionResult[kIdentifierKey] = self.identifier
+        ChoiceQuestionResult[kStartDateKey] = self.startDate
+        ChoiceQuestionResult[kEndDateKey] = self.endDate
+        ChoiceQuestionResult[QuestionResultUserInfoKey] = self.userInfo
+        if let answer = self.jsonSerializedAnswer() {
+            ChoiceQuestionResult[answer.key] = answer.value
```
syoung-smallwisdom: to keep compatibility with surveys and existing implementations, should probably include both `answer.key` and "answer"

BridgeAppSDK/ORKResult+SBAExtension.swift
syoung-smallwisdom: I have a pull request to add to this PR (confusing, I know). There is a method in it called `SBAJSONObjectForObject`. You could also use that.

BridgeAppSDK/ORKResult+SBAExtension.swift
Erin-Mounts: It doesn't yet, actually, but it probably makes sense to move this there anyway. I put it here partly because BridgeSDK doesn't actually use this internally, but mostly because I was in a hurry.

BridgeAppSDK/ORKResult+SBAExtension.swift
Erin-Mounts, line 251: yep, I just copied & pasted this from AppCore and adjusted for Swift syntax.

BridgeAppSDK/ORKResult+SBAExtension.swift
Erin-Mounts: yes it builds, 'override' is because I didn't remove that qualifier after discovering that didn't actually work in Swift (will do so now), and jsonObject() is the right method name, I think, because it returns a JSON-safe dictionary, not just a dictionary. It can't override jsonObject() implementations in superclasses, but it can still implement the SBAJSONObject protocol itself.

BridgeAppSDK/ORKResult+SBAExtension.swift
Erin-Mounts: oops

Erin-Mounts:
I think I've addressed your comments...

BridgeAppSDK/DataArchiving/SBAActivityArchive.swift
```diff
@@ -0,0 +1,68 @@
...
+//
+//  Created by Erin Mounts on 5/18/16.
+//  Copyright © 2016 Sage Bionetworks. All rights reserved.
+//
+
+private let kSurveyCreatedOnKey               = "surveyCreatedOn";
+private let kSurveyGuidKey                    = "surveyGuid";
+private let kSchemaRevisionKey                = "schemaRevision";
+
+class SBAActivityArchive: SBADataArchive {
```
syoung-smallwisdom, line 13: Should be public.

BridgeAppSDK/DataArchiving/SBAActivityArchive.swift
```diff
@@ -0,0 +1,68 @@
...
+    
+    init(reference: String, result: SBAActivityResult) {
```
syoung-smallwisdom, line 15: public

BridgeAppSDK/DataArchiving/SBAActivityArchive.swift
```diff
@@ -0,0 +1,68 @@
...
+        super.init(reference: reference)
+        
+        if (result.schemaRevision != nil) {
```
syoung-smallwisdom, line 18: result schema is never nil.

BridgeAppSDK/DataArchiving/SBAActivityArchive.swift
```diff
@@ -0,0 +1,68 @@
...
+            self.setArchiveInfoObject(result.schemaRevision, forKey: kSchemaRevisionKey)
+        }
+        if let surveyReference = result.schedule.activity.survey {
+            // Survey schema is better matched by created date and survey guid
+            self.setArchiveInfoObject(surveyReference.guid, forKey: kSurveyGuidKey)
+            self.setArchiveInfoObject(surveyReference.createdOn.ISO8601String(), forKey: kSurveyCreatedOnKey)
+        }
```
syoung-smallwisdom, line 25: Also include the taskIdentifier (if applicable) and the SBBScheduledActivity.guid value so that the session and task (training vs. regular) can be tracked.

BridgeAppSDK/DataArchiving/SBAActivityArchive.swift
```diff
@@ -0,0 +1,68 @@
...
+                            
+                            if let urlResult = archiveableResult.result as? NSURL {
+                                archive.insertURLIntoArchive(urlResult, fileName: archiveableResult.filename)
+                            } else if let dictResult = archiveableResult.result as? [NSObject : AnyObject] {
+                                archive.insertDictionaryIntoArchive(dictResult, filename: archiveableResult.filename)
+                            } else if let dataResult = archiveableResult.result as? NSData {
+                                archive.insertDataIntoArchive(dataResult, filename: archiveableResult.filename)
+                            } else {
+                                let className = NSStringFromClass(archiveableResult.result.classForCoder)
+                                fatalError("Unsupported archiveable result type: \(className)")
```
syoung-smallwisdom, line 51: Can we use an assert here instead?

BridgeAppSDK/DataArchiving/SBAActivityArchive.swift
```diff
@@ -0,0 +1,68 @@
...
+                            }
+                        }
+                    }
+                }
+                
+                if let error = archive.completeArchive() {
```
syoung-smallwisdom, line 57: Isn't the error already logged?

BridgeAppSDK/DataArchiving/SBADataArchive.m
```diff
@@ -0,0 +1,265 @@
...
+    NSError *error = nil;
+    if (self.filesList.count) {
+        [self.infoDict setObject:self.filesList forKey:kFilesKey];
+        
+        [self.infoDict setObject:self.reference forKey:kItemKey];
+        [self.infoDict setObject:[[NSBundle mainBundle] appName] forKey:kAppNameKey];
+        [self.infoDict setObject:[self.class appVersion] forKey:kAppVersionKey];
+        [self.infoDict setObject:[[UIDevice currentDevice] deviceInfo] forKey:kPhoneInfoKey];
+
+        // TODO: syoung 04/26/2016 Implement using setArchiveInfoObject: to add these values
```
syoung-smallwisdom, line 194: Remove TODO comment?

BridgeAppSDK/DataArchiving/SBAEncryption.m
```diff
@@ -0,0 +1,211 @@
...
+    }
+    
+    return self;
+}
+
+- (void)encryptFileAtURL:(NSURL *)url withCompletion:(void (^)(NSURL *url, NSError *error))completion
+{
+    NSError * encryptionError = nil;
+    NSData * unencryptedZipData = [NSData dataWithContentsOfFile:url.relativePath];
+    NSData * encryptedZipData = [self.class cmsEncrypt:unencryptedZipData identityPath:[SBAEncryptionHelper pemPath] error:&encryptionError];
```
syoung-smallwisdom, line 63: What happens if the pem file is missing? I think that results in a rather obscure crash. Might be nice to either allow a nil case (for testing) and/or use an NSAssert with more explanation?

BridgeAppSDK/DataArchiving/SBAEncryption.m
```diff
@@ -0,0 +1,211 @@
...
+        BIO_free(tbio);
+    
+    return returnValue;
+}
+
+#pragma mark - helpers
+
++ (NSString*) pemPath
+{
+    return [SBAEncryptionHelper pemPath];
```
syoung-smallwisdom, line 183: Directly accessed above. This code is never called, is it?

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,329 @@
...
+        super.init()
+    }
+}
+
+protocol BridgeUploadableData {
+    // returns result object, result type, and filename
+    func bridgeData(stepIdentifier: String) -> ArchiveableResult?
+}
+
+extension ORKResult: BridgeUploadableData {
```
syoung-smallwisdom, line 98: Should make this public so that apps that have their own implementation of an ORKResult subclass can be archived.

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,329 @@
...
+    }
+}
+
+protocol ORKQuestionResultAnswerJSON {
+    func jsonSerializedAnswer() -> AnswerKeyAndValue?
+}
+
+private var _timeOnlyFormatter: NSDateFormatter? = nil
+
+extension ORKQuestionResult: ORKQuestionResultAnswerJSON {
```
syoung-smallwisdom, line 289: public

BridgeAppSDK/ORKResult+SBAExtension.swift
```diff
@@ -0,0 +1,329 @@
...
+        self.value = value
+        super.init()
+    }
+}
+
+protocol ORKQuestionResultAnswerJSON {
+    func jsonSerializedAnswer() -> AnswerKeyAndValue?
+}
+
+private var _timeOnlyFormatter: NSDateFormatter? = nil
```
syoung-smallwisdom, line 287: Is this still used?

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -80,6 +80,11 @@ public protocol SBABridgeInfo: class {
      * Mapping of task identifier and associated info for creating a task
      */
     var taskMap: [NSDictionary]? { get }
+    
+    /**
+     * Name of .pem certificate file to use for uploading to Bridge (without the .pem extension)
+     */
+    var certificateName: String? { get }
```
syoung-smallwisdom, line 87: If this is defined as optional then should have a drop-through for the encryption.

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -133,6 +138,10 @@ public class SBABridgeInfoPList : NSObject, SBABridgeInfo {
     public var schemaMap: [NSDictionary]? {
         return self.plist["schemaMapping"] as? [NSDictionary]
     }
+    
+    public var certificateName: String? {
+        return self.plist["certificateName"] as? String
```
syoung-smallwisdom, line 143: Make non-optional and return studyIdentifier if nil?

BridgeAppSDK/SBAConsentSignature.swift
```diff
@@ -33,7 +33,7 @@
 
 import Foundation
 
-public protocol SBAConsentSignature : class, NSSecureCoding {
+@objc public protocol SBAConsentSignature : class, NSSecureCoding {
```
syoung-smallwisdom, line 36: Does it need this?

syoung-smallwisdom:
I have a few comments but for some reason they are not showing up here. Mostly about public vs. internal. Tired tonight. Will look over and merge in the morning when my brain is working better. :)

syoung-smallwisdom:
Pulling in as-is. Only issues are small enough that they can be handled in refactor.

PR #48, opened 2016-05-18
Cleanup and fix to NSArray and NSDictionary implementation of jsonObject method

syoung-smallwisdom:
Found a bug in the implementation of the NSArray and NSDictionary methods for jsonObject, plus the method with the key mapping was unnecessary with the current implementation.

I didn't want to push this to the shared branch b/c openssl requires about 1/2 hour to build. Therefore, I did not build and run unit tests, but thought this might be useful.

### DIFF
--- a/BridgeAppSDK/SBAJSONObject.m
+++ b/BridgeAppSDK/SBAJSONObject.m
@@ -34,18 +34,6 @@
 #import "SBAJSONObject.h"
 @import BridgeSDK;
 
-@implementation NSObject (SBAJSONObject)
-
-- (id)jsonObject {
-    if ([self respondsToSelector:@selector(dictionaryRepresentation)]) {
-        return [[(id)self dictionaryRepresentation] jsonObject];
-    }
-    NSAssert(@"jsonObject method not implemented for class %@", NSStringFromClass([self class]));
-    return @{};
-}
-
-@end
-
 @implementation NSString (SBAJSONObject)
 
 - (id)jsonObject {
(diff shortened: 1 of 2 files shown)


BridgeAppSDK/SBAJSONObject.m
```diff
@@ -197,40 +185,40 @@ - (id)jsonObject {
 
 @end
 
-id sba_JSONObjectForObject(id object, NSString * key, NSDictionary <NSString *, NSFormatter *> *formatterMap) {
-    if ([object respondsToSelector:@selector(jsonObjectWithFormatterMap:)]) {
-        return [object jsonObjectWithFormatterMap:formatterMap];
+id SBAJSONObjectForObject(id object) {
+    if (![object isKindOfClass:[NSObject class]]){
+        // If this is not an object that inherits from NSObject class then return NSNull
+        return [NSNull null];
```
Erin-Mounts: I'd like to either have this assert rather than just silently returning NSNull, or have the method signature take an NSObject, as it indicates a programming error.

BridgeAppSDK/SBAJSONObject.m
```diff
@@ -197,40 +185,40 @@ - (id)jsonObject {
...
-            return [dictionary jsonObjectWithFormatterMap:formatterMap];
+            return [dictionary jsonObject];
         }
     }
-    else if ([object respondsToSelector:@selector(jsonObjectWithFormatter:)]) {
-        return [object jsonObjectWithFormatter:formatterMap[key]];
-    }
     else {
+        // Finally, drop through to the description
         return [object description];
```
Erin-Mounts, line 206: Assert would be good here, too.

BridgeAppSDK/SBAJSONObject.m
Erin-Mounts: Alternatively, it could call a Swift function that would see if it was a Swift non-class object (Dictionary, Array, etc.) that it could convert to JSON and assert/return NSNull if it failed.

Erin-Mounts:
looks good

PR #52, opened 2016-05-19
Consolidate timing step results into single result

syoung-smallwisdom:


BridgeAppSDK/ORKFormStep+Result.swift
```diff
@@ -0,0 +1,163 @@
+//
+//  ORKFormStep+ResultCreation.swift
```
Erin-Mounts, line 2: filename changed?

Erin-Mounts:
looks good

PR #54, opened 2016-05-20
Fix identifier for frequency step result

Erin-Mounts:


syoung-smallwisdom:
whoops. 
Fixing now as well as a different issue discovered when looking at the schema for mPower. Closing in favor of https://github.com/Sage-Bionetworks/BridgeAppSDK/pull/55

PR #55, opened 2016-05-20
Fix identifier for added selection result and use custom result type to allow key override

syoung-smallwisdom:
mPower already has an implementation that assumes the schema is setup with "items" as the keyword for the selected medication rather than "choiceAnswers".


Erin-Mounts:
looks good

PR #56, opened 2016-05-20
Fix subtask step results to return mutated results

syoung-smallwisdom:


BridgeAppSDK/SBATrackedDataResult.swift
```diff
@@ -37,6 +37,10 @@ public class SBATrackedDataSelectionResult: ORKQuestionResult {
     
     public var selectedItems: [SBATrackedDataObject]?
     
+    override init() {
```
Erin-Mounts, line 40: Is this needed?

Erin-Mounts:
One question, otherwise looks good to go

BridgeAppSDK/SBATrackedDataResult.swift
syoung-smallwisdom, line 40: Yes, with swift, any init that you do not override is no longer a valid initializer (throws) and therefore, when copyWithZone instantiates using [[[self class] alloc] init] you get a runtime exception.

PR #58, opened 2016-05-26
Fixes to learn more and optionally do not require name/signature for consent.

syoung-smallwisdom:


BridgeAppSDK/Resources/SBAResourceFinder.swift
```diff
@@ -82,12 +82,41 @@ public class SBAResourceFinder: NSObject {
...
     
+    func importHTML(input: String) -> String {
+        
+        // setup string
+        var html = input
+        func search(str: String, _ range: Range<String.Index>?) -> Range<String.Index>? {
+            return html.rangeOfString(str, options: .CaseInsensitiveSearch, range: range, locale: nil)
+        }
+        
+        // search for <import href="resourceName.html" /> and replace with contents of resource file
```
Erin-Mounts, line 100: does iOS still not support <link rel="import" href="...">? (see e.g. http://www.html5rocks.com/en/tutorials/webcomponents/imports/)

Erin-Mounts:
one question, otherwise good

BridgeAppSDK/Resources/SBAResourceFinder.swift
syoung-smallwisdom, line 100: Because this is designed for override, the replacement html may or may _not_ be in the same bundle as the original html. This was a way that I could replace the contact and IRB text with the appropriate text for Lux without have to include a full translation for all the html files that use it.

Erin-Mounts, line 100: Ah, makes sense.

PR #59, opened 2016-05-26
Added external ID registration step

syoung-smallwisdom:
This is still somewhat WIP but will _work_ for Lilly and mPower-Lux

### DIFF
--- a/BridgeAppSDK/SBAExternalIDOnboardingController.swift
+++ b/BridgeAppSDK/SBAExternalIDOnboardingController.swift
@@ -45,6 +45,31 @@ public protocol SBAExternalIDOnboardingController: class, SBASharedInfoControlle
     func goNext()
 }
 
+public protocol SBAExternalIDOnboardingControllerWithVerification: SBAExternalIDOnboardingController {
+    
+    var registrationStep: SBAExternalIDRegistrationStep? { get }
+    
+    func goInitialStep()
+    func goVerificationStep()
+    
+}
+
+extension SBAExternalIDOnboardingControllerWithVerification {
+    
+    func setupInitialEntry() {
+        self.goInitialStep()
+        self.registrationCodeTextField.text = nil
+        self.registrationCodeTextField.becomeFirstResponder()
+    }
+    
+    func setupVerification() {
+        self.goVerificationStep()
+        self.registrationCodeTextField.text = nil
+        self.registrationCodeTextField.becomeFirstResponder()
+    }
+    
+}
+
 public extension SBAExternalIDOnboardingController {
     
     public func handleViewDidAppear() {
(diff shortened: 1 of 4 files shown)


BridgeAppSDK/SBAExternalIDOnboardingController.swift
```diff
@@ -53,22 +78,46 @@ public extension SBAExternalIDOnboardingController {
...
+        guard let text: String = self.registrationCodeTextField.text?.trim()
             where text.characters.count > 0 else {
-                let message = NSLocalizedString("Please enter a valid registration code.", comment: "Message for invalid registration code")
+                let message = Localization.localizedString("SBA_REGISTRATION_INVALID_CODE")
                 showAlertWithOk(nil, message: message, actionHandler: nil)
                 return
         }
         
+        // Check if this is a step view controller with verification
+        if let stepVC = self as? SBAExternalIDOnboardingControllerWithVerification,
```
Erin-Mounts, line 90: don't like an object checking its own type at run-time, maybe put this in a method defined in the base protocol that can be overridden by the derived protocol (and returns true/false so the caller knows whether to exit early)?

BridgeAppSDK/SBAExternalIDOnboardingController.swift
```diff
@@ -53,22 +78,46 @@ public extension SBAExternalIDOnboardingController {
...
         
         sharedUser.loginUser(externalId: text) { [weak self] error in
             if let error = error {
                 self?.hideLoadingView({
-                    let title = NSLocalizedString("Registration Failed", comment: "Title for error when registration fails")
+                    let title = Localization.localizedString("SBA_REGISTRATION_FAILED_TITLE")
                     let message = error.localizedBridgeErrorMessage
-                    self?.showAlertWithOk(title, message: message, actionHandler: nil)
+                    self?.showAlertWithOk(title, message: message, actionHandler: { (_) in
+                        (self as? SBAExternalIDOnboardingControllerWithVerification)?.setupInitialEntry()
```
Erin-Mounts, line 118: see above re:checking one's own type at runtime

Erin-Mounts:
one nit/pet-peeve, otherwise looks good

BridgeAppSDK/SBAExternalIDOnboardingController.swift
syoung-smallwisdom, line 90: Swift Protocol extensions don't work that way. You can't override a function. It won't get called (only the base protocol implementation will be called.) My goal was to allow reverse-compatibility with Lilly so that Lilly doesn't have to implement methods it doesn't use (optional methods - which aren't allowed in swift)

Erin-Mounts, line 90: figured it might be something like that--fair enough.

PR #61, opened 2016-06-07
Fix which identifier is used for creating a task from a dictionary.

syoung-smallwisdom:
Note: Currently this code is only used by Lilly, and the Study Drug Tracker task was the only implementor of this path. The result was that the schema identifier was not being correctly mapped for upload to the server.


Erin-Mounts:
looks good

PR #62, opened 2016-06-08
Bridge 1382 - Allow activities when no network

Erin-Mounts:
requires https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/79


syoung-smallwisdom:
looks good

PR #66, opened 2016-06-16
Do not refresh the schedule until after callback that the schedule has been updated

syoung-smallwisdom:
…


Erin-Mounts:
looks good

PR #68, opened 2016-06-23
Bridge 1390

Erin-Mounts:
Implement handleEventsForBackgroundURLSession app delegate method


syoung-smallwisdom:
Looks good. 

Note: submodules will need to be updated again with merged commit but I'm merging now b/c the event handling needs to be added.

PR #70, opened 2016-06-30
Archiving fixes

syoung-smallwisdom:
Issues found and fixed:
1. If the walking task is skipped, then the behavior was inconsistent. By adding a "skip" result, this could be tracked _or_ the behavior of the base class can be overridden to remove this `SBAActivityResult`.
2. Fixed nil `choiceAnswers` for the medication timing where each med is tracked.
3. Do not include an `SBAActivityResult` if none of the included steps have results.


Erin-Mounts:
looks good

PR #72, opened 2016-06-30
Add placeholder text for html content

syoung-smallwisdom:
Otherwise ResearchKit will throw an exception during test.


Erin-Mounts:
looks good

PR #74, opened 2016-07-11
Fix the start/end dates for the split out tasks and add these values to metadata

syoung-smallwisdom:
…


BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -422,10 +422,21 @@ public class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORK
         guard taskViewController.result.results != nil else { return [] }
         
         let taskResult = taskViewController.result
+        
+        func outputDate(inputDate: NSDate?, comparison:NSComparisonResult) -> NSDate {
```
Erin-Mounts, line 430: The purpose of this function is kind of opaque--can you explain with a comment?

Erin-Mounts:
looks good

PR #76, opened 2016-07-13
BRIDGE-1303: fix Motion Activity permission button

Erin-Mounts:
so it doesn’t show “Allow” when permissions have already been granted on the phone, when registering with a fresh install.

### DIFF
--- a/BridgeAppSDK/SBAPermissionsManager.m
+++ b/BridgeAppSDK/SBAPermissionsManager.m
@@ -490,7 +490,20 @@ NSString *const kHKWorkoutTypeKey           = @"HKWorkoutType";
 #if TARGET_IPHONE_SIMULATOR
     return YES;
 #else
-    return self.coreMotionPermissionStatus == SBAPermissionStatusAuthorized;
+    BOOL permission;
+    
+    // available iOS 9 and later: a way to check motion sensor recorder permissions without explicitly requesting them.
+    // TODO: emm 2016-07-12 redo the whole permissions thing so this situation doesn't even come up. We should
+    // be requesting permissions either ahead of time or as we first need them, but not asking the user for
+    // permission to even *ask* them for the permissions, which winds up with not being able to set them in Settings
+    // later if they don't allow us to ask the first time through.
+    if ([CMSensorRecorder class]) {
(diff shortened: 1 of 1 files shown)


BridgeAppSDK/SBAPermissionsManager.m
```diff
@@ -490,7 +490,20 @@ - (BOOL)isPermissionsGrantedForCoreMotion {
...
 #else
-    return self.coreMotionPermissionStatus == SBAPermissionStatusAuthorized;
+    BOOL permission;
+    
+    // available iOS 9 and later: a way to check motion sensor recorder permissions without explicitly requesting them.
+    // TODO: emm 2016-07-12 redo the whole permissions thing so this situation doesn't even come up. We should
+    // be requesting permissions either ahead of time or as we first need them, but not asking the user for
+    // permission to even *ask* them for the permissions, which winds up with not being able to set them in Settings
+    // later if they don't allow us to ask the first time through.
+    if ([CMSensorRecorder class]) {
```
syoung-smallwisdom, line 500: CMSensorRecorder doesn't exist in iOS 8?  Did you test this with an iOS 8 device/sim?

syoung-smallwisdom, line 500:
```
id sensorRecordClass = NSClassFromString(@"CMSensorRecorder");
if (sensorRecorderClass) {
permission = [sensorRecordClass isAuthorizedForRecording];
....
```

Erin-Mounts, line 500: The class doesn't exist in iOS 8 but since we're linking against iOS SDK 9 it knows about it and this is the recommended way to check for the availability of weakly-linked classes in iOS:

http://stackoverflow.com/questions/3057325/weak-linking-check-if-a-class-exists-and-use-that-class/10068079#10068079
https://developer.apple.com/library/ios/documentation/DeveloperTools/Conceptual/cross_development/Using/using.html#//apple_ref/doc/uid/20002000-SW3

syoung-smallwisdom, line 500: Groovy. Thanks for checking. :)

syoung-smallwisdom:
looks good

PR #78, opened 2016-07-14
Tracking bug: fix for breaking change to RK/master

syoung-smallwisdom:
There is a change to how navigation is handled in RK/master that mutated step results (that were not the step results returned by the stepViewController) are overwritten. Short-term solution is to move the mutation to the split results code.


Erin-Mounts:
looks good

PR #81, opened 2016-07-21
Refactor the activity timing for tracking each item

syoung-smallwisdom:
This converts the activity timing steps to use a subclass of ORKFormStep if tracking grouped and a subclass of ORKPageStep if tracking each. While the ORKStep and ORKTask cannot mutate the ORKTaskResult, the ORKStepViewController can mutate the ORKStepResult for it's step. This implementation takes advantage of that to override the step result returned by the view controller to add the consolidated result that is returned to the server (and read into the synapse table). I am hoping to also refactor the selection/frequency steps to use an ORKPageStep subclass, but this one is a little less complicated so thought I would start with this part.

This requires pointing at Sage-Bionetworks/ResearchKit/sage-master until such time as https://github.com/ResearchKit/ResearchKit/pull/758 is accepted into ResearchKit/master and we have had a chance to fix any breaking changes in Apple's master branch.


Erin-Mounts:
Looks good

PR #82, opened 2016-07-25
Refactor tracked data selection and frequency

syoung-smallwisdom:
Refactored the tracked data selection step to use a grouped step for selection/frequency.


Erin-Mounts:
looks good

PR #84, opened 2016-07-29
Refactor external ID handling and add first pass of permissions step

syoung-smallwisdom:
Note: ResearchKit already has handling for asking users for required permissions for activities where the step requires them. The UX is not awesome and this step could _potentially_ be designed to display in an activity flow to explain why the request is being made.

I went with a simplified approach to the external ID where it _always_ asks for confirmation.

I also went with a simplified approach to the permissions where it will iterate through all the permissions where the user has not granted authorization and return an error indicating that changes require going through the settings.


BridgeAppSDK/SBANavigableOrderedTask.swift
```diff
@@ -34,13 +34,24 @@
...
- * Currently defined usage is to allow the SBANavigableOrderedTask to check if a step has a navigation rule.
+ Define the navigation rule as a protocol to allow for protocol-oriented extention (multiple inheritance).
+ Currently defined usage is to allow the SBANavigableOrderedTask to check if a step has a navigation rule.
  */
 public protocol SBANavigationRule: class, NSSecureCoding {
     func nextStepIdentifier(result: ORKTaskResult, additionalTaskResults:[ORKTaskResult]?) -> String?
 }
 
+/**
+ A naviation skip rule applies to this step to allow that step to be skipped.
```
Erin-Mounts: navi_g_ation

BridgeAppSDK/SBAPermissionsStep.swift
```diff
@@ -0,0 +1,214 @@
+//
+//  SBAPermissionsTask.swift
```
Erin-Mounts: SBAPermissions_Step_

Erin-Mounts:
couple of nits about typos, otherwise looks good, feel free to merge

PR #85, opened 2016-08-10
Survey scheduling

syoung-smallwisdom:
TODO: Verify the archive and uploading.

This includes both changes to externalID/permissions as a result of manual UI testing and changes to the surveys to use ORKQuestionStep (which apparently looks different from ORKFormStep).  I have not verified archive and uploading yet.


Erin-Mounts:
Looks good

PR #87, opened 2016-08-11
BRIDGE-1403 Consolidate duplicate step results

syoung-smallwisdom:
The voice activity uses ORKNavigationTask which includes duplicate results if the room is too loud. Consolidate the results into a single step.


Erin-Mounts:
looks good

PR #88, opened 2016-08-16
BRIDGE-1391 Modified the permissions step to allow for failing the task

syoung-smallwisdom:
If the permissions step is not optional and the given permissions are not all supported then the step will send a failed step message to the delegate.


Erin-Mounts:
looks good

PR #89, opened 2016-08-17
Bridge-1438 Add mood survey to BridgeAppSDK

syoung-smallwisdom:
This survey has been added to RK/sage-master and there is an outstanding pull request https://github.com/ResearchKit/ResearchKit/pull/793 associated with it.

### DIFF
--- a/BridgeAppSDK/SBAActiveTask.swift
+++ b/BridgeAppSDK/SBAActiveTask.swift
@@ -43,16 +43,18 @@ public enum SBAActiveTaskType {
     case voice
     case walking
     case tremor
+    case moodSurvey
     
     init(name: String?) {
         guard let type = name else { self = .custom(nil); return }
         switch(type) {
-        case "tapping"  : self = .tapping
-        case "memory"   : self = .memory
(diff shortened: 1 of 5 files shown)


BridgeAppSDK/SBAActiveTask.swift
```diff
@@ -97,6 +99,16 @@ extension ORKTremorActiveTaskOption {
     }
 }
 
+extension ORKMoodSurveyFrequency {
```
Erin-Mounts, line 102: Does RK have to know something about the scheduling of the mood survey?

Erin-Mounts:
Question about the purpose of ORKMoodSurveyFrequency, otherwise looks good

BridgeAppSDK/SBAActiveTask.swift
syoung-smallwisdom, line 102: This enum is passed into RK to determine language for the questions. Either "today, I feel" or "this week, I feel".

Basically, I am imagining that there might be a case for only checking in once a week and wanted to support that.

PR #93, opened 2016-08-31
Strip out all but the last audio file.

syoung-smallwisdom:


syoung-smallwisdom:
Including with Smart4Sure PR

PR #94, opened 2016-08-31
Changes to support Smart4SURE build

syoung-smallwisdom:


BridgeAppSDK/DataArchiving/SBADataArchive.m
```diff
@@ -163,6 +163,12 @@ - (void)insertDictionaryIntoArchive:(NSDictionary *)dictionary filename:(NSStrin
 
 - (void)insertDataIntoArchive :(NSData *)data filename: (NSString *)filename
 {
+    // Check that the file has not already been added
```
Erin-Mounts, line 166: For voice tests, it's probably not as straightforward but I think it would be better to include the last sound check file rather than the first, as that's going to be the one that corresponds to the time/place the voice test was finally recorded, in case researchers want to use it to try to compensate for the background noise in their analyses. I'm not sure this always applies in general to all possible file duplication scenarios going forward, though. Eventually we might want to be able to specify in a schema that there might be more than one of a given file, and use some naming convention when archiving them, so that researchers can see all the data and decide what to make of it, but for now this is the only situation I know about where we actually might care which of the "duplicate" files we save.

BridgeAppSDK/Localization.swift
```diff
@@ -154,92 +156,6 @@ public class Localization: NSObject {
     
     // MARK: List of the strings used in this module.
     public class var listOfAllLocalizedStrings: [String : String] {
-        return listOfMyLocalizedStrings
+        return [:]
```
Erin-Mounts: Should we maybe just do away with this method and go with the .strings file in all the modules that subclass this (e.g. mPowerSDK, various apps)? If so, would be better to do so now before anyone else sees/follows this pattern...

BridgeAppSDK/NSDate+Utilities.swift
```diff
@@ -41,5 +41,13 @@ extension NSDate {
         let components = calendar.components(unitFlags, fromDate: self)
         return calendar.dateFromComponents(components) ?? self
     }
+    
+    var isToday: Bool {
+        return self.startOfDay() == NSDate().startOfDay()
+    }
+    
+    var isTomorrow: Bool {
+        return self.startOfDay() == NSDate(timeIntervalSinceNow: 24 * 60 * 60).startOfDay()
```
Erin-Mounts: We've done this in a number of places already (1 day == 86400 seconds), but it turns out it isn't safe; it's wrong for an hour on changes between Standard and Daylight time and back (and for 1 second if there's a leap second, but that's hardly a problem and pretty rare anyway). I think the "right" way is to use date components and add one day. We should fix these in our code as we find them.

BridgeAppSDK/NSPredicate+Utilities.swift
```diff
@@ -36,9 +36,17 @@ import Foundation
...
+    public convenience init(week: NSDate, dateKey: String) {
+        self.init(date: week, dateKey: dateKey, numberOfDays: 7)
+    }
+    
+    public convenience init(date: NSDate, dateKey: String, numberOfDays: Double) {
         let calendar = NSCalendar.currentCalendar()
-        let start = calendar.startOfDayForDate(day)
-        let end = start.dateByAddingTimeInterval(24*60*60)
+        let start = calendar.startOfDayForDate(date)
+        let end = start.dateByAddingTimeInterval(numberOfDays * 24 * 60 * 60)
```
Erin-Mounts: see previous comment regarding length of day and standard vs DST

BridgeAppSDK/ORKCollectionResult+SBAExtensions.m
```diff
@@ -78,32 +78,26 @@ @implementation ORKTaskResult (SBAExtensions)
...
-                        [stepResults addObjectsFromArray:dupResults];
-                    }
-                }
-                stepResult.results = stepResults;
-            }
-            [results addObject:stepResult];
+    NSMutableArray *identifiers = [NSMutableArray new];
+    NSEnumerator *enumerator = [self.results reverseObjectEnumerator];
+    for (ORKStepResult *stepResult in enumerator) {
+        if (![identifiers containsObject:stepResult.identifier]) {
```
Erin-Mounts, line 97: see previous comment about using the latest rather than first (for now)

BridgeAppSDK/SBACheckmarkView.swift
```diff
@@ -0,0 +1,135 @@
...
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import UIKit
+
+class SBACheckmarkView: UIView {
```
Erin-Mounts, line 36: nice

BridgeAppSDK/SBADirectNavigationRule.swift
```diff
@@ -0,0 +1,55 @@
...
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import Foundation
+
+/**
+ * The direct navigation rule allows for a step to be displayed with a direct
+ * pointer to something other than the next step in the sequencial order defined by
```
Erin-Mounts: sequential

BridgeAppSDK/SBAInstructionStep.swift
```diff
@@ -97,14 +81,24 @@ public final class SBADirectNavigationStep: ORKInstructionStep, SBADirectNavigat
         self.customTypeIdentifier = customTypeIdentifier
     }
     
+    public override func stepViewControllerClass() -> AnyClass {
+        if self.isCompletionStep && self.image == nil {
```
Erin-Mounts, line 88: purpose of this logic isn't entirely clear--can you add an explanatory comment?

BridgeAppSDK/SBANewsFeedItem.h
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2015, Apple Inc. All rights reserved.
```
Erin-Mounts: update copyright? or did you just copy the file without changes?

BridgeAppSDK/SBANewsFeedItem.m
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) 2015, Apple Inc. All rights reserved.
```
Erin-Mounts: update unless copied unchanged

BridgeAppSDK/SBANewsFeedManager.h
```diff
@@ -0,0 +1,58 @@
+//
+//  SBANewsFeedManager.h
+//  APCAppCore
```
Erin-Mounts: update

BridgeAppSDK/SBANewsFeedManager.m
```diff
@@ -0,0 +1,143 @@
+//
+//  SBANewsFeedManager.m
+//  APCAppCore
```
Erin-Mounts: update

BridgeAppSDK/SBANewsFeedParser.h
```diff
@@ -1,8 +1,4 @@
-//
-//  SBATaskViewController.h
-//  BridgeAppSDK
-//
-//  Copyright © 2016 Sage Bionetworks. All rights reserved.
+// Copyright (c) 2015, Apple Inc. All rights reserved.
```
Erin-Mounts: probably Git is being weird thinking these were ever the same file, but does copyright need updating here or was the file copied unchanged?

BridgeAppSDK/SBANewsFeedParser.m
```diff
@@ -0,0 +1,197 @@
+// Copyright (c) 2015, Apple Inc. All rights reserved.
```
Erin-Mounts: update unless file copied unchanged

BridgeAppSDK/DataArchiving/SBADataArchive.m
syoung-smallwisdom: This is a belt and suspenders check. There were two different bugs that were causing the issue and they have been resolved in a way that is appropriate to the error condition. In this case, it is saying "in a production app, pass the data in such a way that it can be collected by the service. in debug, throw an assert."

BridgeAppSDK/Localization.swift
syoung-smallwisdom: Good point. I can change that thought it means updating Lilly.

BridgeAppSDK/DataArchiving/SBADataArchive.m
Erin-Mounts, line 166: OK, makes sense.

BridgeAppSDK/ORKCollectionResult+SBAExtensions.m
syoung-smallwisdom: Using the reverse enumeration which is why, below, it uses `-insertObject:atIndex:`

Erin-Mounts, line 97: Ah, didn’t notice that, perfect.

BridgeAppSDK/SBATextChoice.swift
```diff
@@ -42,7 +42,7 @@ public protocol SBATextChoice  {
 
 extension SBATextChoice {
     func createORKTextChoice() -> ORKTextChoice {
-        return ORKTextChoice(text: self.choiceText, detailText: self.choiceDetail, value: self.choiceValue, exclusive: self.exclusive)
+        return ORKTextChoice(text: self.choiceText.trim() ?? "", detailText: self.choiceDetail?.trim(), value: self.choiceValue, exclusive: self.exclusive)
```
Erin-Mounts, line 45: should choiceDetail also fall back to ""?

BridgeAppSDK/SBBScheduledActivity+Filters.m
```diff
@@ -70,6 +70,10 @@ + (NSPredicate *) scheduledTomorrowPredicate {
     return [[NSPredicate alloc] initWithDay:[NSDate dateWithTimeIntervalSinceNow:24*60*60] dateKey:NSStringFromSelector(@selector(scheduledOn))];
 }
 
++ (NSPredicate *) scheduledComingUpPredicate: (NSInteger)numberOfDays {
+    return [[NSPredicate alloc] initWithDate:[NSDate dateWithTimeIntervalSinceNow:24*60*60] dateKey:NSStringFromSelector(@selector(scheduledOn)) numberOfDays:numberOfDays];
```
Erin-Mounts: refer to prior comment about assuming 24hr day length

BridgeAppSDK/SBANewsFeedItem.h
syoung-smallwisdom: I changed the name but nothing else which is why I left the original copyright.

BridgeAppSDK/String+Utilities.swift
```diff
@@ -48,4 +48,20 @@ extension String {
         return self.substringFromIndex(range.endIndex)
     }
     
+    public func stringByRemovingNewlineCharacters() -> String {
+        guard let result = (self as NSString).mutableCopy() as? NSMutableString else { return self }
+        let set = NSCharacterSet.newlineCharacterSet()
+        var range = result.rangeOfCharacterFromSet(set)
```
Erin-Mounts: can you do the thing where you split the string on members of the newline character set and then re-join the resulting array with " "? E.g. in obj-C it would be: 
result = [[self componentsSeparatedByCharactersInSet:set] componentsJoinedByString:@" "];
so swift would maybe be something like
self.componentsSeparatedByCharactersInSet(set).componentsJoinedByString(" ")
...?

BridgeAppSDK/SBATextChoice.swift
syoung-smallwisdom: No. only `text` is required. `detailText` can be passed as a `nil`.

BridgeAppSDK/SBANewsFeedManager.h
Erin-Mounts: (I meant APCAppCore --> BridgeAppSDK, but updating copyright is good too)

Erin-Mounts:
looks good, merge when ready

BridgeAppSDK/String+Utilities.swift
syoung-smallwisdom: because I was tired that day. ¯_(ツ)_/¯

PR #95, opened 2016-09-06
BRIDGE-1475 Implement default handling for a catastrophic failure.

syoung-smallwisdom:
Note: The CatastrophicError.storyboard file does not include any strings that require localization.

This allows for the implementation of the app version kill switch with a custom URL for enterprise apps.


BridgeAppSDK.xcodeproj/project.pbxproj
```diff
@@ -2193,7 +2201,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.3;
```
Erin-Mounts, line 2204: Is there a specific reason for changing the deployment target from 8.0?

Erin-Mounts, line 2204: ...or is that at the project level, and you override it specifically for each target?

Erin-Mounts:
question about deployment target settings, otherwise looks good, merge when ready

BridgeAppSDK.xcodeproj/project.pbxproj
syoung-smallwisdom: iOS 10 is coming out this week. Set BridgeAppSDK to minimum deployment of 8.4 and the sample app to min target of 9.3 as per Apple's recommendations for min targets.

PR #96, opened 2016-09-09
build 3 - fix archiving crash

syoung-smallwisdom:
Rev to build 3 and fix the crash I introduced by trying to add an assert in the data archiving. Oops.

Note: I think this will also cause all production builds to _not_ upload properly.


Erin-Mounts:
looks good

PR #101, opened 2016-09-19
Xcode 8 Migration

syoung-smallwisdom:
These are the required changes for Swift 3.0

This has been smoke tested for Lilly and Smart4SURE


BridgeAppSDKTests/SBASurveyFactoryTests.swift
```diff
@@ -1436,7 +1436,7 @@ class SBASurveyFactoryTests: XCTestCase {
         // a step size of 37 is not divisible by 100 so is invalid
```
Erin-Mounts: does this comment no longer quite match the code below?

BridgeAppSDK/SBAUser.swift
```diff
@@ -43,8 +43,8 @@ public final class SBAUser: NSObject, SBAUserWrapper {
     
     let lockQueue = DispatchQueue(label: "org.sagebase.UserLockQueue", attributes: [])
 
-    public func logout() {
-        lockQueue.async {
+    public func resetStoredUserData() {
+        lockQueue.sync {
```
Erin-Mounts: does this really need to be sync? you're not returning a value.

syoung-smallwisdom: Yes, but I do want this to happen before the code goes through setup and checks for what the current "user state" is _or_ on logout. In both of those cases, the data needs to be cleared synchronously. So that in the next step, the correct view controller for the current state will be displayed.

BridgeAppSDKTests/SBASurveyFactoryTests.swift
syoung-smallwisdom: Updated the comment for clarity.

BridgeAppSDK/SBAUser.swift
Erin-Mounts: Doesn't it access the user state via the lockQueue, though, ensuring that this will finish first?

syoung-smallwisdom: Good point... so... why didn't it? Hmmm....

syoung-smallwisdom: Okay... so this is another thing that changed with Swift 3.0
Should be a serial FIFO queue now.

Erin-Mounts:
looks good

PR #107, opened 2016-09-28
Update permissions manager to get the localized permission from bundle.

syoung-smallwisdom:
This will also throw an assert if the required key is missing from the info plist.

### DIFF
--- a/BridgeAppSDK/SBAPermissionsManager.m
+++ b/BridgeAppSDK/SBAPermissionsManager.m
@@ -34,6 +34,7 @@
 #import "SBAPermissionsManager.h"
 #import "SBABridgeAppSDKDelegate.h"
 
+#import <Foundation/Foundation.h>
 #import <UIKit/UIKit.h>
 #import <CoreMotion/CoreMotion.h>
 #import <CoreLocation/CoreLocation.h>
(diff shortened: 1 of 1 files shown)


BridgeAppSDK/SBAPermissionsManager.m
```diff
@@ -237,18 +238,29 @@ - (NSString *)permissionDescriptionForType:(SBAPermissionsType)type {
...
+            return [self bundlePermissionDescriptionForKey: @"NSMotionUsageDescription" defaultValue:@"Using the motion co-processor allows the app to determine your activity, helping the study better understand how activity level may influence disease."];
         case SBAPermissionsTypeMicrophone:
-            return NSLocalizedStringWithDefaultValue(@"SBA_MICROPHONE_PERMISSIONS_DESCRIPTION", nil, SBABundle(), @"Access to microphone is required for your Voice Recording Activity.", @"");
-        case SBAPermissionsTypeCamera:
+            return [self bundlePermissionDescriptionForKey: @"NSMicrophoneUsageDescription" defaultValue:@"Access to microphone is required for your Voice Recording Activity."];
         case SBAPermissionsTypePhotoLibrary:
+            return [self bundlePermissionDescriptionForKey: @"NSPhotoLibraryUsageDescription" defaultValue:@"Photo Library"];
+        case SBAPermissionsTypeCamera:
```
Erin-Mounts, line 248: At least MoleMapper needs access to the camera (though probably gets the permissions outside of Bridge). Does the photo library permission include camera, and we are deprecating a separate camera type? Or...?

syoung-smallwisdom: SBAPermissionsTypePhotoLibrary and  SBAPermissionsTypeCamera were previously undocumented in AppCore with drop-through to "not supported"

Since I found a required key for the photo library, I added that to the list. Left Camera alone b/c I'm not sure what the proper permissions are for it. As long as we are continuing to use the permissions manager in it's current (less than awesome) format, I figured I would add an assert to alert devs that this key needs to be added to the permissions. This issue didn't show up with my previous on-device testing, but not sure if that was b/c I hadn't upgraded my phone from iOS 9 and was relying upon the simulator to catch iOS 10 issues.

Erin-Mounts: Fair enough.

Erin-Mounts:
looks good

PR #113, opened 2016-10-20
Implement custom UI for the eligibility step.

syoung-smallwisdom:
Requires: https://github.com/ResearchKit/ResearchKit/pull/845


BridgeAppSDK/SBAToggleFormStep.swift
```diff
@@ -0,0 +1,170 @@
...
+    public func numberOfRows(inSection section: Int) -> Int {
+        return self.formItems?.count ?? 0
+    }
+    
+    public func reuseIdentifierForRow(at indexPath: IndexPath) -> String {
+        return SBAToggleTableViewCell.reuseIdentifier
+    }
+    
+    let minCellHeight: CGFloat = 120
+    let maxCellHeight: CGFloat = 175
```
Erin-Mounts: Could these be defined as constraints in the storyboard instead of being hardcoded?

Erin-Mounts (approved):
Question about hardcoded size constraints, otherwise looks good, merge when ready

BridgeAppSDK/SBAToggleFormStep.swift
syoung-smallwisdom: Doesn't use a storyboard but should be possible to use IBDesignable properties. I'll look into it.

Erin-Mounts:
looks good

PR #120, opened 2016-10-27
BRIDGE-1562 Fix permissions warnings and localization.

syoung-smallwisdom:
Note: This is not a full refactor of permissions UI/UX. Just cleaning up some of the deprecated permission handling.


Erin-Mounts:
looks good

PR #121, opened 2016-10-28
Fix some nit UI issues

syoung-smallwisdom:


BridgeAppSDK/SBAActivityTableViewCell.swift
```diff
@@ -35,32 +35,10 @@ import UIKit
...
 open class SBAActivityTableViewCell: UITableViewCell {
     
-    @IBOutlet weak var uncheckedView: UIView!
-    @IBOutlet weak var checkmarkImageView: UIImageView!
-    @IBOutlet open weak var titleLabel: UILabel!
-    @IBOutlet open weak var subtitleLabel: UILabel!
-    @IBOutlet open weak var timeLabel: UILabel?
+    @IBOutlet open var titleLabel: UILabel!
+    @IBOutlet open var subtitleLabel: UILabel!
+    @IBOutlet open var timeLabel: UILabel?
```
Erin-Mounts, line 40: shouldn't IBOutlets always be weak? or is Swift smart enough to know that by implication?

Erin-Mounts (approved):
One question about non-weak IBOutlets, otherwise looks good.

BridgeAppSDK/SBAActivityTableViewCell.swift
syoung-smallwisdom: "Best practices" on IBOutlet changes depending upon the day and who you talk to (including Apple). Apple defaults to using weak and ! in swift. This means that if the view is removed, you could end up referencing it even though it is has been deallocated (and is now nil). By removing the weak, this is saying "don't dealloc until I am deallocated" and _not_ that this will cause a memory leak. Therefore, if someone is building this using `initWithFrame:` instead of using a nib or storyboard, then they can instantiate the label without having to use a temporary strong ivar until they add it to the parent view.  That is, do this:

```
self.titleLabel = UILabel()
self.contentView.addSubview(self.titleLabel)
```

rather than this:

```
let titleLabel = UILabel()
self.contentView.addSubview(titleLabel)
self.titleLabel = titleLabel
```

Erin-Mounts: makes sense.

PR #124, opened 2016-11-09
Added public access to storyboard and json file names

dephillipsmichael:
Added public access to storyboard and json file names so that subclass' of SBAOnboardingManager can access them.


syoung-smallwisdom:
Whoops. Thanks.

PR #126, opened 2016-11-15
Added profile image property to SBAUser

dephillipsmichael:
See comments in code

### DIFF
--- a/BridgeAppSDK/SBAUser.swift
+++ b/BridgeAppSDK/SBAUser.swift
@@ -67,6 +67,7 @@ public final class SBAUser: NSObject, SBAUserWrapper {
     let kExternalIdKey = "externalId"
     let kGenderKey = "gender"
     let kBirthdateKey = "birthdate"
+    let kProfileImagePropertyKey = "profileImage"
     
     public var sessionToken: String? {
         get {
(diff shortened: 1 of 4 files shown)


BridgeAppSDK/SBAUser.swift
```diff
@@ -157,6 +158,24 @@ public final class SBAUser: NSObject, SBAUserWrapper {
...
+        get {
+            if let profileImageData = getKeychainObject(kProfileImagePropertyKey) as? Data {
+                return UIImage(data: profileImageData)
+            } else {
+                return nil
+            }
+        }
+        set (newValue) {
+            if let newValueUnwrapped = newValue {
+                let dataValue = UIImageJPEGRepresentation(newValueUnwrapped, 1.0)
```
dephillipsmichael, line 171: I went with UIImageJPEGRepresentation because it is faster and less storage space than lossless UIImagePNGRepresentation.  We could switch the property to be Data? to provide the developer more control over the storage format, but I think this provides an easier interface.

syoung-smallwisdom: Yes, I would agree that a profile image is easier to work with as an image.

BridgeAppSDK/SBAUserWrapper.swift
```diff
@@ -85,6 +85,11 @@ public protocol SBAUserWrapper: class, SBBAuthManagerDelegateProtocol {
     var consentSignature: SBAConsentSignatureWrapper? { get set }
 
     /**
+     * Profile Image is stored in the keychain
+     */
+    var profileImage: UIImage? { get set }
```
syoung-smallwisdom, line 90: Please add this to the unit test mocks as well. Unit tests are broken by this change.

syoung-smallwisdom (changes requested):
Need to fix unit test bundle.

BridgeAppSDK/SBAUserWrapper.swift
dephillipsmichael: Good catch, I'm on it.

dephillipsmichael: Fixed

syoung-smallwisdom (approved):
Looks good

PR #128, opened 2016-11-16
Change account management steps to use protocols

syoung-smallwisdom:
Some of the steps in the account management inherit from ORKQuestionStepViewController or ORKFormStepViewController. Change these to use protocol extensions in order to allow the step view controller to *not* inherit from these view controllers.


Erin-Mounts:
looks good

PR #131, opened 2016-12-01
Bridge 1586 combo activity

syoung-smallwisdom:
Fixes to ensure that the result sets are not mutated by filtering and consolidation methods. These methods are used to split the data and to include navigation handling. They broke when I fixed a bug in ResearchKit where an array could include a pointer to the same object at different positions in the array.


Erin-Mounts:
looks good

PR #133, opened 2016-12-02
Delete RK files when we're done with them

Erin-Mounts:
so they don't keep piling up and using up storage space

### DIFF
--- a/BridgeAppSDK/SBAScheduledActivityManager.swift
+++ b/BridgeAppSDK/SBAScheduledActivityManager.swift
@@ -331,7 +331,8 @@ open class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORKTa
         do {
             try FileManager.default.removeItem(at: outputDirectory)
         } catch let error as NSError {
-            debugPrint("Unable to remove ResearchKit output directory at \(outputDirectory): \(error)")
+            print("Error removing ResearchKit output directory: \(error.localizedFailureReason)")
+            debugPrint("\tat: \(outputDirectory)")
         }
     }
     
--- a/BridgeAppSDK/SBAScheduledActivityManager.swift
+++ b/BridgeAppSDK/SBAScheduledActivityManager.swift
@@ -322,8 +322,22 @@ open class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORKTa
         return activities.find({ $0.taskIdentifier == taskIdentifier })
     }
     
+    /**
+     Delete the output directory.
+     */
+    @objc(deleteOutputDirectoryForTaskViewController:)
+    open func deleteOutputDirectory(for taskViewController: ORKTaskViewController) {
+        guard let outputDirectory = taskViewController.outputDirectory else {return}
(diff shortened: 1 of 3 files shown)


BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -322,8 +322,22 @@ open class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORKTa
...
+    /**
+     Delete the output directory.
+     */
+    @objc(deleteOutputDirectoryForTaskViewController:)
+    open func deleteOutputDirectory(for taskViewController: ORKTaskViewController) {
+        guard let outputDirectory = taskViewController.outputDirectory else {return}
+        do {
+            try FileManager.default.removeItem(at: outputDirectory)
+        } catch let error as NSError {
+            debugPrint("Unable to remove ResearchKit output directory at \(outputDirectory): \(error)")
```
syoung-smallwisdom: Won't this mean that in the release build that the error is silenced?

BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -347,7 +361,38 @@ open class SBAScheduledActivityManager: NSObject, SBASharedInfoController, ORKTa
             taskViewController.task?.resetTrackedDataChanges()
         }
         
-        taskViewController.dismiss(animated: true) {}
+        taskViewController.dismiss(animated: true) {
+            self.offMainQueue.async {
+                self.deleteOutputDirectory(for: taskViewController)
+                #if DEBUG
+                DispatchQueue.main.async {
```
syoung-smallwisdom, line 369: why going back on main thread?

BridgeAppSDK/SBAScheduledActivityManager.swift
Erin-Mounts: True. I'm never sure where the right place is to draw the line between "things we might need to troubleshoot in production apps" and "things we want to catch before it goes to production but don't want to clutter the logs with/potentially reveal sensitive info/etc. in production apps".

BridgeAppSDK/SBAScheduledActivityManager.swift
Erin-Mounts: I've had issues in the past with logging to the console from random threads leaving some debug output interleaved with other output, making it confusing and hard to follow. Not sure it would be a problem here, or that this would even actually help with that if it were though, so... yeah. You want I should remove it?

syoung-smallwisdom: right, that makes sense. I was just curious. thanks.

BridgeAppSDK/SBAScheduledActivityManager.swift
syoung-smallwisdom: What about splitting into two lines: if debug print the directory, if not then just print the error reason?

syoung-smallwisdom (approved):

PR #135, opened 2016-12-08
Bridge 1514 HealthKit and stub out for demographic data archive

syoung-smallwisdom:
Updates to profile info to allow for archiving and unit tests. 

Also includes WIP for demographic archive uploader.

### DIFF
--- a/BridgeAppSDK/SBAProfileInfoForm.swift
+++ b/BridgeAppSDK/SBAProfileInfoForm.swift
@@ -40,7 +40,6 @@ public enum SBAProfileInfoOption : String {
     case name                   = "name"
     case birthdate              = "birthdate"
     case gender                 = "gender"
-    case biologicalSex          = "biologicalSex"
     case bloodType              = "bloodType"
     case fitzpatrickSkinType    = "fitzpatrickSkinType"
     case wheelchairUse          = "wheelchairUse"
@@ -179,10 +178,6 @@ public struct SBAProfileInfoOptions {
                 formItems.append(formItem)
                 
             case .gender:
-                let formItem = makeGenderFormItem(with: option.rawValue)
-                formItems.append(formItem)
-                
-            case .biologicalSex:
                 let formItem = makeGenderFormItem(with: option.rawValue)
                 formItems.append(formItem)
                 
--- a/BridgeAppSDK/SBAProfileInfoForm.swift
+++ b/BridgeAppSDK/SBAProfileInfoForm.swift
@@ -44,6 +44,10 @@ public enum SBAProfileInfoOption : String {
     case bloodType              = "bloodType"
     case fitzpatrickSkinType    = "fitzpatrickSkinType"
     case wheelchairUse          = "wheelchairUse"
+    case height                 = "height"
+    case weight                 = "weight"
+    case wakeTime               = "wakeTime"
+    case sleepTime              = "sleepTime"
 }
 
 enum SBAProfileInfoOptionsError: Error {
@@ -149,11 +153,11 @@ public struct SBAProfileInfoOptions {
             switch option {
                 
             case .email:
-                let formItem = makeEmailFormItem(option)
+                let formItem = makeEmailFormItem(option.rawValue)
                 formItems.append(formItem)
             
             case .password:
-                let (formItem, answerFormat) = makePasswordFormItem(option)
+                let (formItem, answerFormat) = makePasswordFormItem(option.rawValue)
                 formItems.append(formItem)
                 
                 // confirmation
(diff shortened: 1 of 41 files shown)


BridgeAppSDK/SBAProfileInfoForm.swift
```diff
@@ -40,9 +40,14 @@ public enum SBAProfileInfoOption : String {
     case name                   = "name"
     case birthdate              = "birthdate"
     case gender                 = "gender"
+    case biologicalSex          = "biologicalSex"
```
Erin-Mounts: separate fields for biologicalSex and gender? HealthKit AFAIK only has the one and calls it biologicalSex, which is scientifically meaningless, and in apps we usually call it gender, for convoluted social reasons, although we use the HK biologicalSex value for it when we have HK permissions. All of which is to say, we don't currently distinguish these or even define them really, so do we need both?

BridgeAppSDK/SBAProfileInfoForm.swift
```diff
@@ -140,78 +145,107 @@ public struct SBAProfileInfoOptions {
...
+                let formItem = makeBirthdateFormItem(with: option.rawValue)
                 formItems.append(formItem)
                 
             case .gender:
-                let formItem = makeGenderFormItem(option)
+                let formItem = makeGenderFormItem(with: option.rawValue)
+                formItems.append(formItem)
+                
+            case .biologicalSex:
+                let formItem = makeGenderFormItem(with: option.rawValue)
```
Erin-Mounts: case in point re: my other comment.

BridgeAppSDK/SBAProfileInfoForm.swift
```diff
@@ -330,28 +364,95 @@ public struct SBAProfileInfoOptions {
...
-        let formItem = ORKFormItem(identifier: option.rawValue,
+        let formItem = ORKFormItem(identifier: identifier,
                                    text: Localization.localizedString("WHEELCHAIR_USE_FORM_ITEM_TITLE"),
                                    answerFormat: answerFormat,
                                    optional: false)
         
         return formItem
     }
     
+    func makeHeightFormItem(with identifier: String) -> ORKFormItem {
```
Erin-Mounts, line 383: Will this work for entering heights in feet & inches? I.e. does the ORKFormItem magically support that when the units are feet?

Unrelated, why do we even still use those ridiculous units? /rant

BridgeAppSDK/SBARegistrationStep.swift
```diff
@@ -128,6 +128,9 @@ extension SBARegistrationStepController {
                 if let gender = self!.gender {
                     self!.sharedUser.gender = gender
                 }
+                if let biologicalSex = self!.biologicalSex {
```
Erin-Mounts: goes with the prior comments re: gender/biologicalSex

BridgeAppSDK/SBAUser.swift
```diff
@@ -66,6 +66,7 @@ public final class SBAUser: NSObject, SBAUserWrapper {
     let kConsentSignatureKey = "ConsentSignature"
     let kExternalIdKey = "externalId"
     let kGenderKey = "gender"
+    let kBiologicalSexKey = "biologicalSex"
```
Erin-Mounts: again, do we want/need both

BridgeAppSDK/SBAUser.swift
```diff
@@ -123,6 +124,19 @@ public final class SBAUser: NSObject, SBAUserWrapper {
         }
     }
     
+    public var biologicalSex: HKBiologicalSex {
+        get {
+            // syoung 11/17/2016 Older versions of AppCore apps use the gender key and do not differentiate
+            // between gender identity and biological sex (which is defined as sex at birth).
+            // In medical research, a person's biological sex is usually more relevant to the study.
```
Erin-Mounts: In medical research, if you ask for this info it's always as a rough proxy for something else, something specific you actually do care about, like the presence or absence of certain organs prone to certain cancers, the relative concentrations and/or fluctuations of various steroidal hormones, or occasionally the chromosomal karyotype, and probably a couple of other things I'm forgetting. Better to ask for the specific info that's important for your purposes than rely on a term which doesn't perfectly map to any of them, let alone all of them at once.

BridgeAppSDK/SBAUserWrapper.swift
```diff
@@ -65,11 +65,16 @@ public protocol SBAUserWrapper: class, SBBAuthManagerDelegateProtocol {
     var password: String? { get set }
     
     /**
-     * Gender is stored in the keychain
+     * Gender describes the characteristics that a society or culture delineates as masculine or feminine. Stored in the keychain.
```
Erin-Mounts: That's one possible way of defining it, but if that's what we're asking, shouldn't the choices be  "masculine/feminine/other" rather than "male/female/other?" Mainly though, it's simply used in place of "sex" which some people are squeamish about saying.

BridgeAppSDK/SBAUserWrapper.swift
```diff
@@ -65,11 +65,16 @@ public protocol SBAUserWrapper: class, SBBAuthManagerDelegateProtocol {
...
      */
     var gender: HKBiologicalSex { get set }
     
     /**
+     * Biological sex refers to the biological sex of the person at birth. Stored in the keychain.
```
Erin-Mounts: "Sex" isn't actually any more medically meaningful than "gender" as it turns out (refer to my previous comments for elaboration).

BridgeAppSDKSample/ProfileTableViewController.swift
```diff
@@ -46,27 +46,19 @@ class ProfileTableViewController: UITableViewController, SBASharedInfoController
...
-    }
-    
     override func viewDidLoad() {
         super.viewDidLoad()
         
         let version = Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString")
         versionLabel.text = "\(Localization.localizedAppName) \(version!), build \(Bundle.main.appVersion())"
         appLabel.text = Localization.localizedAppName
-        
-        // TODO: syoung 10/20/2016 Enable pause/resume study. For now, just hide the option.
```
Erin-Mounts, line 68: for now this is fine but I fear eventually we're going to have to bring this back; the consent quiz specifically asks whether you're allowed to pause/resume the study at any time (the correct answer is "yes").

BridgeAppSDKTests/MockObjects/MockUser.m
```diff
@@ -50,6 +50,7 @@ @implementation MockUser
 @synthesize dataSharingScope;
 @synthesize onboardingStepIdentifier;
 @synthesize gender;
+@synthesize biologicalSex;
```
Erin-Mounts: ditto

Erin-Mounts (commented):
I apologize in advance for the extensive discourse on gender and sex. Also one or two comments are about other unrelated stuff.

BridgeAppSDK/SBAProfileInfoForm.swift
syoung-smallwisdom: This came up with Woody's UI for one of the apps. (Don't remember which one) which referred to "sex at birth" (or something like that) and a second question about gender identity. Apparently, in most cases, the "sex at birth" is more medically relevant to the study while sometimes "gender identity" is more relevant for a particular study.

BridgeAppSDK/SBAProfileInfoForm.swift
syoung-smallwisdom: Yes, somewhere along the way, it was added to ResearchKit.

BridgeAppSDKSample/ProfileTableViewController.swift
syoung-smallwisdom: True, but I would rather remove that question from the sample app than *add* to this one.

syoung-smallwisdom:
WRT gender vs. biological sex, the differentiation came about with going over the UI with Woody and asking about some of the questions he was asking in gathering demographic data. I can follow up but I figured that since I was working in this area, I'd go ahead and add it.

In coming up with naming conventions, I used google. Specifically,
https://en.wikipedia.org/wiki/Sex_and_gender_distinction
http://www.med.monash.edu.au/gendermed/sexandgender.html

BridgeAppSDK/SBAUserWrapper.swift
syoung-smallwisdom: Good point, though I would prefer to let the actual language of the questions wait for Woody and Christine's input and for now just use the terms defined in ResearchKit.

Erin-Mounts (approved):
Looks good.

PR #137, opened 2016-12-09
BRIDGE-1626 preload surveys and do not mark survey as complete if not available.

syoung-smallwisdom:
…


Erin-Mounts (approved):
Looks good

PR #138, opened 2016-12-09
BRIDGE-1391 Use optional flag for checking if the task should dismiss…

syoung-smallwisdom:
… w/o permissions.

This will show the skip step for optional health kit permission step, but this is actually "okay" from a basic UI flow point of view. (Not very pretty but... *shrug*)


Erin-Mounts:
looks good

PR #139, opened 2016-12-09
Feature/make more methods accessible

EricSiegNW:
Add open to handleWrongEmailAction and handleResendEmail action so we can call these explicitly from sub classes. In our case we want to tie them to button actions specific to our UI rather than the alert approach used in the BridgeAppSDK.


syoung-smallwisdom:
looks good

PR #140, opened 2016-12-13
Create singletons for user and bridge info

syoung-smallwisdom:


Erin-Mounts:
Looks good, actually.

PR #144, opened 2016-12-15
Add keychain to BridgeAppSDK that can set access group and service.

syoung-smallwisdom:
Actually setting up for shared app access to the keychain is complicated. We will want to create documentation to walk new developers who are less comfortable with Apple iOS through the process since they need to include the App ID Prefix from the developer portal.

I chose to move the code into BridgeAppSDK rather than trying to modify either the keychain wrapper in BridgeSDK or ResearchKit because the BridgeSDK wasn’t working (not sure why) and I didn’t want to rely upon the ResearchKit version which might change in ways we cannot control.

Note: This is a direct copy/paste of the ResearchKit version (kept for legacy apps) that was ported from AppCore with RK v1.3. It includes *not* using the access group if running in the simulator so  I am not certain how to make this work with a keyboard extension that is being testing using the simulator.

In order to implement this you will need to do the following:

1. http://evgenii.com/blog/sharing-keychain-in-ios/ has instructions for how to *get* the string for the access group.

2. Once you have the string for your access group, you will want to add the following keys to the BridgeInfo-private.plist file (You do not want to include the App ID Prefix in an open source file):

````
"keychainService" = <shared name for the service>
"keychainAccessGroup" = <access group as defined in the link above>
````

### DIFF
--- a/BridgeAppSDK/SBAKeychainWrapper.m
+++ b/BridgeAppSDK/SBAKeychainWrapper.m
@@ -1,0 +1,317 @@
+ /**
+ Copyright (c) 2015, Apple Inc. All rights reserved. 
+ 
+ Redistribution and use in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ 
+ 1.  Redistributions of source code must retain the above copyright notice, this
+ list of conditions and the following disclaimer.
+ 
+ 2.  Redistributions in binary form must reproduce the above copyright notice, 
+ this list of conditions and the following disclaimer in the documentation and/or 
+ other materials provided with the distribution. 
+ 
+ 3.  Neither the name of the copyright holder(s) nor the names of any contributors 
+ may be used to endorse or promote products derived from this software without 
+ specific prior written permission. No license is granted to the trademarks of 
+ the copyright holders even if such marks are included in this software. 
+ 
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" 
+ AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+ IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+ ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE 
+ FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL 
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR 
+ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: BridgeAppSDK/SBAKeychainWrapper.m [hunk 1]
```diff
@@ -0,0 +1,317 @@
...
+            }
+        } else if (status == errSecItemNotFound) {
+            NSMutableDictionary *attributes = [[NSMutableDictionary alloc] init];
+            [attributes setObject:(__bridge id)kSecClassGenericPassword forKey:(__bridge id)kSecClass];
+            [attributes setObject:service forKey:(__bridge id)kSecAttrService];
+            [attributes setObject:key forKey:(__bridge id)kSecAttrAccount];
+#if TARGET_OS_IPHONE || (defined(MAC_OS_X_VERSION_10_9) && MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_9)
+            [attributes setObject:(__bridge id)kSecAttrAccessibleAfterFirstUnlock forKey:(__bridge id)kSecAttrAccessible];
+#endif
+            [attributes setObject:data forKey:(__bridge id)kSecValueData];
```
Erin-Mounts: I'm assuming this is copypasted from RK, so if this is actually a bug it's a bug in RK as well, but... if data is nil, in the case where the item already existed, it would just remove it with no further comment, but if it didn't already exist, it will try to add the item with nil data, and presumably end up failing with a message that it encountered an error trying to add the item. Just double-checking that this is the desired behavior in both cases, or at least, that maintaining parity with the RK behavior is more of a concern.

BridgeAppSDK/SBAKeychainWrapper.m
```diff
@@ -0,0 +1,317 @@
...
+            if (status != errSecSuccess) {
+                if (error) {
+                    *error = [NSError errorWithDomain:NSOSStatusErrorDomain
+                                                 code:status
+                                             userInfo:@{NSLocalizedDescriptionKey: [Localization localizedString:@"KEYCHAIN_ADD_ERROR_MESSAGE"]}];
+                }
+                returnValue = NO;
+            }
+        } else {
+            returnValue = NO;
```
Erin-Mounts, line 217: Again, guessing it's the same in RK, but if it finds the item with no error, it attempts to modify it, and hands back a descriptive error object if it fails; if the error is "item not found", it attempts to create it, again handing back an error object with details in case of problems; but if it's anything else, it just returns NO with no explanation. In other words, it seems like it would make sense to pass back the error details here as well, no? Or could this break something that depends on the current RK behavior of not setting the error value here?

BridgeAppSDK/SBAKeychainWrapper.m
```diff
@@ -0,0 +1,317 @@
...
+        OSStatus status = SecItemDelete((__bridge CFDictionaryRef)itemToDelete);
+        if (status != errSecSuccess) {
+            if (error) {
+                *error = [NSError errorWithDomain:NSOSStatusErrorDomain
+                                             code:status
+                                         userInfo:@{NSLocalizedDescriptionKey: [Localization localizedString:@"KEYCHAIN_DELETE_ERROR_MESSAGE"]}];
+            }
+            returnValue = NO;
+        } else {
+            returnValue = YES;
```
Erin-Mounts: Once again, presumably from RK like this, but won't this just end up returning YES or NO based on the status from the last item in the list, regardless of the success or failure of removal of all the previous items? And is this what we want?

Erin-Mounts (approved):
A few comments/questions about behavior that's presumably the same here as in the ResearchKit keychain wrapper, otherwise looks fine. Merge when ready.

PR #146, opened 2016-12-20
Cleanup and fixes

syoung-smallwisdom:
Change BridgeInfo to use days ahead/behind for caching.
Add method for overriding the default survey factory.


syoung-smallwisdom:
Pushing changes

PR #147, opened 2016-12-21
Fix unit tests

syoung-smallwisdom:


syoung-smallwisdom:
Fixed issues that were keeping BiAffect from archiving a test flight build.

PR #148, opened 2016-12-27
Archive Issue

EricSiegNW:
Note: Don't merge this into master. This PR is created to highlight the issue so that Erin/Shannon can pull the branch from my fork to see the issue.

We've struggled with an archive related bug on the Resilience app for a little while now. It will build fine, deploy to phone or simulator fine, etc. However, as soon as we try to archive it we get errors like those seen in the screenshot here: https://www.dropbox.com/s/b6gb4h48l9bo86f/Screenshot%202016-12-09%2014.55.12.png?dl=0

Further research into the issue appears to create a lot of ambiguity as to what the problem might be. There are a lot of "X fixed it for us" issues, but those fixes either don't resolve the issue or are already correct in our project.

In order to experiment further, I wanted to try using the bridging header in the BridgeSampleApp in BridgeAppSDK to see if it would work correctly or also have issues. Since the bridging header exists in the sample app but isn't currently being used, I dropped some references in there along with some code to make sure it gets hit.

After doing this, I'm hitting the same issue in the sample app where it will build/deploy fine but has errors when trying to archive. Would love some help either identifying the issue or identifying a different approach we can take to accessing RK classes comfortably in our project using BridgeAppSDK without hitting this issue. Please also feel free to look at Resilience, in particular the RP classes like RPTextField along with the CreateAccountViewController to see how we're trying to tie into events while modifying the UI to understand our goal.

Here's a SS of the current archive bug being thrown from the sample app in this branch: https://www.dropbox.com/s/86rf4c4256d1wz7/Screenshot%202016-12-27%2013.12.49.png?dl=0

### DIFF
--- a/BridgeAppSDK.xcodeproj/project.pbxproj
+++ b/BridgeAppSDK.xcodeproj/project.pbxproj
@@ -7,6 +7,15 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		04AFE0161E11E11D00BC3C2A /* CustomTextField.swift in Sources */ = {isa = PBXBuildFile; fileRef = 04AFE0151E11E11D00BC3C2A /* CustomTextField.swift */; };
+		04AFE0691E130CB200BC3C2A /* ResearchKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FFC1609F1CFE672100C29AF7 /* ResearchKit.framework */; };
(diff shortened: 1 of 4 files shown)


syoung-smallwisdom:
@EricSiegNW 
I ran into this same issue with trying to use private/internal ResearchKit methods in our app. Unfortunately, the only solution I was able to come to was to submit a PR to ResearchKit to publicly expose the RK methods. I had the same issue where it would run just fine until I tried to archive it. Then archiving would fail.

syoung-smallwisdom:
closing due to inactivity.

PR #149, opened 2016-12-28
BRIDGE-1637 Support multiple skip-to identifiers for SBBSurveyRules 

syoung-smallwisdom:
Note: This is only applicable to SBBSurveyRules. Surveys built using a dictionary still only support a single identifier.


Erin-Mounts:
looks good

PR #153, opened 2017-01-06
add activity label to metadata.json for activities

Erin-Mounts:
makes it easier for data analysts to map task results in Synapse to which schedule they are from


syoung-smallwisdom (approved):
looks good

PR #155, opened 2017-01-09
Fix height/weight profile form items to NOT ask healthkit permission

syoung-smallwisdom:


Erin-Mounts (commented):
Are we not going to try to store these in/fetch them from HealthKit, then?

syoung-smallwisdom:
This is a work-around. ResearchKit has an answer format type that handles height, weight, etc. for use in a *profile* (where the height/weight are assumed to be that of an adult user). By using their version, we can have RK handle getting the data from HealthKit (if permission is granted). However, by default, ResearchKit will throw up a permission screen (it's awful UI - shows up as soon as the task view controller is displayed) so I added a flag to the `ORKHealthKitQuantityTypeAnswerFormat`  to keep this from happening. But when I added height/weight (in a previous pull request) I forget to set the flag. Didn't realize it b/c simulator was remembering a previous install of the app that had already granted that permission.

syoung-smallwisdom:
This is not intended to address adding new "types" for a child's height/weight and blood pressure.

Erin-Mounts:
ok looks good then

PR #157, opened 2017-01-12
Revert to full scheduled activity guid in archive

Erin-Mounts:
Lilly was using both parts, and it’s possible (at least in Lilly, not necessarily in general) but tricky to infer the originally scheduled date/time in Synapse otherwise. Also the “scheduleIdentifier” is unique to the schedule but not to the particular scheduledActivity instance, so using the scheduledActivityGuid key name for the schedule identifier is misleading.

That being said, it’s also useful (probably usually more so) to have the scheduleIdentifier and scheduledOn available separately with no parsing to split them out during analysis, so I’ve added two new keys to the metadata for those.


syoung-smallwisdom (approved):
looks good

PR #158, opened 2017-01-14
Bridge 1651 factory submodule

syoung-smallwisdom:
Requires: https://github.com/Sage-Bionetworks/ResearchUXFactory-iOS/pull/1
Requires: https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/109

Ready for review. The weirdness I discovered on Friday turned out to be unrelated to this pull request. See https://sagebionetworks.jira.com/browse/BRIDGE-1653

There are some changes to how data groups are managed but otherwise, the changes herein are to allow pulling out the UX factory into its own SDK. Note: I am using the same `SBA` prefix for both libraries. This is intentional to limit any refactoring that must be done by clients that are currently using BridgeAppSDK.


BridgeAppSDK/SBAAppExtensionSharedInfoController.swift
```diff
@@ -36,7 +36,7 @@ import UIKit
 /**
  The `SBAAppExtensionSharedInfoController` is a light-weight object that can be used to 
  manage the user and bridge info for an app extension. This class assumes that the current
- user and bridge info use the shared singletons defined on `SBAUser` and `SBABridgeInfoPList`,
+ user and bridge info use the shared singletons defined on `SBAUser` and `SBABridgeInfoManager`,
```
Erin-Mounts: Should this be SBAInfoManager?

BridgeAppSDK/SBABridgeManager.m
```diff
@@ -33,15 +33,54 @@
...
 + (void)setupWithStudy:(NSString *)study
         cacheDaysAhead:(NSInteger)cacheDaysAhead
        cacheDaysBehind:(NSInteger)cacheDaysBehind
            environment:(SBBEnvironment)environment
           authDelegate:(id <SBBAuthManagerDelegateProtocol>) authDelegate {
     
+    [self addResourceBundleIfNeeded];
+    
+    // If the auth delegate is also the current user then set that
+    if ([authDelegate conformsToProtocol:@protocol(SBAParticipantInfo)]) {
```
Erin-Mounts, line 79: seems like an odd place to be checking/setting this, albeit probably convenient

BridgeAppSDK/SBAConsentDocumentFactory.swift
```diff
@@ -32,91 +32,12 @@
 //
 
 import ResearchKit
```
Erin-Mounts, line 34: Do we want to eventually refactor away BridgeAppSDK's direct dependency on ResearchKit, relying on ResearchUXFactory to wrap the underlying implementation details for us and allowing us to potentially "plug in" other UX implementations someday? Not a current design goal but maybe something to consider? Or maybe it doesn't even make sense. Just thinking on the screen here.

BridgeAppSDK/SBABridgeManager.m
syoung-smallwisdom: This is a deprecated method but it's also the method wedMD is accessing directly b/c they already have an app where the delegate is obj-c so it can't inherit from `SBAAppDelegate`.

Basically, once I'd pulled out the factory into its own library, I needed a way to point at the `SBAParticipantInfo`. Figured this was best approach since then webMD would still work.

BridgeAppSDK/SBAConsentDocumentFactory.swift
syoung-smallwisdom: No.  Only reason for pulling out ResearchUXFactory into it's own SDK is to untie from BridgeSDK (because for whatever reason it too confusing to use the SBB model objects and not the connection to the server.)  BridgeAppSDK is still intended to be dependent on ResearchKit and implements steps that are subclasses of `ORKStep`.

BridgeAppSDK/SBABridgeManager.m
Erin-Mounts: ok, now it makes perfect sense

BridgeAppSDK/SBAConsentDocumentFactory.swift
Erin-Mounts: ok, in that case looks good, merge when ready...

PR #159, opened 2017-01-17
BRIDGE-1653 Only update the data groups if the response was successful.

syoung-smallwisdom:
Requires: https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/108


Erin-Mounts:
looks good

PR #161, opened 2017-01-20
Bridge 1655 participant caching

Erin-Mounts:
Requires https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/110

Remove obsolete/deprecated delegate method, support password(forAuthManager:) being required now, update to BridgeSDK with fix for Bridge 1655

### DIFF
--- a/BridgeAppSDK/SBAUserWrapper+Bridge.swift
+++ b/BridgeAppSDK/SBAUserWrapper+Bridge.swift
@@ -232,7 +232,7 @@ public extension SBAUserWrapper {
      @param completion  Completion handler
      */
     public func verifyRegistration(_ completion: ((Error?) -> Void)?) {
-        guard let username = self.email?(forAuthManager: nil), let password = self.password(forAuthManager: nil) else {
+        guard let username = self.email(forAuthManager: nil), let password = self.password(forAuthManager: nil) else {
             assertionFailure("Attempting to login without a stored username and password")
             return
         }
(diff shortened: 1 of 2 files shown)


BridgeAppSDK/SBAUserWrapper+Bridge.swift
```diff
@@ -232,7 +232,7 @@ public extension SBAUserWrapper {
      @param completion  Completion handler
      */
     public func verifyRegistration(_ completion: ((Error?) -> Void)?) {
-        guard let username = self.email?(forAuthManager: nil), let password = self.password?(forAuthManager: nil) else {
+        guard let username = self.email?(forAuthManager: nil), let password = self.password(forAuthManager: nil) else {
```
syoung-smallwisdom: email looks like it still has the `?`. Is it still optional?

Erin-Mounts: yes, I left it optional. I can't imagine not wanting it, an argument could be easily made for requiring it, but it's always been optional and nothing about its actual technical necessity has changed.

syoung-smallwisdom: That's kind of confusing. Why would password be required but not email? Don't you need both to authenticate?

Erin-Mounts:
now Requires https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/111

syoung-smallwisdom (approved):
Looks good. merge when ready.

PR #163, opened 2017-01-25
Bridge 1508 documentation refactor

syoung-smallwisdom:
Update to match refactored rules and move some of the functionality into the shared base.


Erin-Mounts (approved):
looks good

PR #164, opened 2017-01-27
Move archiving to bridgesdk

Erin-Mounts:
requires https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/112


BridgeAppSDK/DataArchiving/SBADataArchive.h
```diff
@@ -1,135 +0,0 @@
...
-// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
-// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
-// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
-// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
-// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-//
-
-#import <Foundation/Foundation.h>
```
syoung-smallwisdom, line 35: Please include this class `deprecated` as an inheriting class from `SBBDataArchive` so that anyone who is currently inheriting from it has a trail to follow on fixing their deprecation warnings. Thanks.

BridgeAppSDK/SBABridgeManager.m
```diff
@@ -63,7 +63,7 @@ + (void)setupWithBridgeInfo:(id <SBBBridgeInfoProtocol>)bridgeInfo
     [SBAInfoManager sharedManager].currentParticipant = participant;
     
     // Setup the Bridge study
-    [BridgeSDK setupWithBridgeInfo:bridgeInfo];
+    [BridgeSDK setupWithBridgeInfo:bridgeInfo errorUIDelegate:SBAAppDelegate.shared];
```
syoung-smallwisdom: Assumes that this is setup with `SBAAppDelegate` which is not necessarily the case. Because obj-c cannot inherit from Swift 3.0 classes (dangit), apps that were created with an obj-c based app delegate must implement this stuff themselves. (At least for now until adding an obj-c shared delegate bubbles up in priority).

Would be better to be explicit with the setup like we are doing with the auth delegate.

syoung-smallwisdom (approved):
looks good.

PR #165, opened 2017-01-31
Add methods to set the navigation bar tint color appearance

syoung-smallwisdom:
Requires: https://github.com/Sage-Bionetworks/ResearchUXFactory-iOS/pull/6

Added nav bar tint color to the default methods. This allows for setting the task navigation bar to some color other than white.  Since the way ResearchKit wants you to do this is obscure, I thought it would be helpful to add as an example.


Erin-Mounts (approved):
Looks good, merge when ready

PR #176, opened 2017-02-09
Bridge 1701

EricSiegNW:
Add the reason for consent to review step to populate alert, ResearchKit doesn't appear to be pulling it from the consent document and instead just has a property on the step


syoung-smallwisdom (changes requested):
Could you please do this instead?

````
            if let consentReview = inputItem as? SBAConsentReviewOptions,
                consentReview.usesDeprecatedOnboarding {
                // If this uses the deprecated onboarding (consent review defined by ORKConsentReviewStep)
                // then return the step from the superclass.
                return super.createConsentStep(inputItem: inputItem, subtype: subtype)
            }
            else {
                let review = inputItem as! SBAFormStepSurveyItem
                let step = SBAConsentReviewStep(inputItem: review, inDocument: self.consentDocument, factory: self)
                return step;
            }
````

syoung-smallwisdom (approved):

syoung-smallwisdom:
Forgot that this was pinned to a build before adding the factory SDK.

looks good.

PR #177, opened 2017-02-09
Return superclass step for certain consent steps.

syoung-smallwisdom:
Fixes a bug found by Eric and fixed in the mPower 1.4 branch


Erin-Mounts (approved):
looks good

Issue #183, opened 2017-02-14
Request: Maintain consistent question font size when UI hints present


webmd-lsetch:
Issue: When I add UI hints to questions, the font size for the question triples in size. In one question, this is causing the list of drugs scroll space to become really small. See attached screen shots for differences in font sizes. 
![bp-question](https://cloud.githubusercontent.com/assets/25513532/22933090/6e673d7e-f298-11e6-8c66-23fc4ef7e973.png)
![weight-question](https://cloud.githubusercontent.com/assets/25513532/22933089/6e647ddc-f298-11e6-85db-5d7e2f4f1764.png)

syoung-smallwisdom:
@webmd-lsetch 
BridgeAppSDK has been updated to make the placement of the prompt text consistent. Changing the font size requires changing ResearchKit and is outside the scope of what we have time to help you with. (sorry) In order to change font size, you will need to look at RK and submit an issue to that repo. Once you have a PR ready, if you ping me on the ResearchKit PR, I will review and merge into our fork. Thanks!

PR #184, opened 2017-02-14
Fix issue where prompt can be placed in either title or text

syoung-smallwisdom:
Partial fix for https://github.com/Sage-Bionetworks/BridgeAppSDK/issues/183

This places the prompt consistently in the same field for each question. It does not fix the issue of font size. That is beyond the scope of what I have time to work on right now.

Also fixes errors in the unit tests introduced with changes to the BridgeSDK in a previous commit.


Erin-Mounts (approved):
looks good

PR #190, opened 2017-03-02
Bridge 1754

EricSiegNW:
tweak language in consent htmls to fix issue with broken IRB reference

### DIFF
--- a/BridgeAppSDKSample/Base.lproj/consent_19sharing_rsch.html
+++ b/BridgeAppSDKSample/Base.lproj/consent_19sharing_rsch.html
@@ -17,7 +17,7 @@ Note that, we may be required by law to share both your contact information and 
 
 <ul style="padding-left:20px;">
 <li>The US Department of Health and Human Services (HHS), the Office for Human Research Protection (OHRP), the Food and Drug Administration (FDA), and other agencies for review of our research procedures.</li>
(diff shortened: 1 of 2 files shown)


BridgeAppSDKSample/Base.lproj/consent_19sharing_rsch.html
```diff
@@ -17,7 +17,7 @@
 
 <ul style="padding-left:20px;">
 <li>The US Department of Health and Human Services (HHS), the Office for Human Research Protection (OHRP), the Food and Drug Administration (FDA), and other agencies for review of our research procedures.</li>
-<li><import href="consent_irb.html" /> so they can monitor the safety, effectiveness, and conduct of our research.</li>
+<li>The Institutional Review Board (IRB) so they can monitor the safety, effectiveness, and conduct of our research.</li>
```
syoung-smallwisdom, line 20: Um? Why would you need to change the sample project consent documents? Also, this is here b/c when we were doing the Luxembourg project, they have a different IRB.

syoung-smallwisdom:
closing due to inactivity

Erin-Mounts:
actually I think the released version 1.4 was built with the branch this pull request was from, so we ought to get it into the Sage fork one way or another.

syoung-smallwisdom:
The change *shouldn't* be necessary and should *not* be changed in the sample app.

PR #191, opened 2017-03-03
update submodule commit

Erin-Mounts:
pull in BridgeSDK fix for sharing getting reset to none when setting data groups


Erin-Mounts:
Looks like some tests are failing because travis CI is running them in GMT time zone and ISO8601 string isn't matching (...Z vs ...+0000). Going to go ahead and merge this for WebMD.

PR #199, opened 2017-04-09
refactor fastlane ci builds

zaro0508:
Refactor CI to make archive builds work.

* update to use fastlane match for code signing
* set fastlane to download correct certificates and profiles
* setup pipelines in build.sh script
* setup fastlane defaults in Scanfile (for testing) and Matchfile (for code signing)
* user bundler to manage fastlane dependencies
* disable xcode automatic manage signing


syoung-smallwisdom:
So... what does this do? And if I am running using Xcode, can I still run the unit tests and debug the sample app?

zaro0508:
I guess your asking specifically about commit 02bd9b3 ?   That commit disables the auto code signing in xcode.  There doesn't seem to be any other way to make command line archive builds work otherwise.  This is the thing I wanted to talk to you and Erin about.  To answer your specific question.. yes, you should be able to debug and run unit tests in xcode.  The only difference is that xcode will be setup to manual code sign and configured to use the shared "developer" profile/certificate that I have setup on the apple portal and ios-certificates repo.

syoung-smallwisdom:
Aren't we already doing that for releases? And yes, I noticed that it isn't the handy-dandy Xcode automated thang, but that's just getting a certificate key, right?

syoung-smallwisdom (approved):

Erin-Mounts (approved):

PR #212, opened 2017-05-09
Work-around for limits on fetching all schedules

syoung-smallwisdom:


BridgeAppSDK/SBABridgeManager.m
```diff
@@ -228,6 +228,67 @@ + (void)fetchChangesToScheduledActivities:(NSArray <SBBScheduledActivity *> *)sc
...
+                    
+                    // Add from this call
+                    [allSchedules addObjectsFromArray:bList];
+                    
+                    // leave the group
+                    dispatch_group_leave(historyGroup);
+                }];
+            }
+            
+            dispatch_group_async(historyGroup, dispatch_get_main_queue(), ^{
```
Erin-Mounts: Should be using dispatch_group_notify() here, I think? The network completion handlers are executed on the NSURLSession.sharedSession's serial NSOperationQueue, which is not the main queue, so this block will execute after the others have all been started, I think, but not necessarily after they've completed.

Erin-Mounts (approved):
looks good

BridgeAppSDK/SBABridgeManager.m
syoung-smallwisdom: Whoops, thanks. Auto complete failed me and unit tests worked b/c I was throwing to the main queue.

PR #215, opened 2017-05-10
Profile support--preliminary

Erin-Mounts:
SBAProfileManager and SBAProfileItem (implementing support for Keychain and UserDefaults profile items) are complete enough to be useful, and tested.  SBAProfileSection and SBAProfileDataSource are still works in progress.


BridgeAppSDK/SBAProfileDataSource.swift
```diff
@@ -0,0 +1,73 @@
...
+     */
+    @objc(profileItemAtIndexPath:)
+    func profileItem(at indexPath: IndexPath) -> SBAProfileItem?
+    
+    /**
+     Called when a row is selected.
+     @param indexPath   The index path for the profile item.
+     */
+    @objc(didSelectRowAtIndexPath:)
+    optional func didSelectRow(at indexPath: IndexPath)
```
syoung-smallwisdom, line 64: Might be nice to include in the comment what is expected. Is this a "show the appropriate view controller" or is it more of a "change the model" kind of a thing?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+import Foundation
+import ResearchUXFactory
+
+@objc
+public protocol SBAProfileItem: NSObjectProtocol {
+    /**
+     key is used to access a specific profile item, and so must be unique across all SBAProfileItems
+     within an app.
+     */
+    var key: String { get }
```
syoung-smallwisdom: This runs the risk of a namespace conflict since `key` is a pretty common variable name. Perhaps `profileKey`?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+    /**
+     demographicKey is the profile item's key in the demographic data upload schema. By default it will
+     be the same as key, but can be different if needed.
+     */
+    var demographicKey: String { get }
+    
+    /**
+     itemType specifies what type to store the profileItem as.
+     */
+    var itemType: String { get }
```
syoung-smallwisdom: Could you use an enum instead? Map the enum to a string? You can use `NS_ENUM_STRING` in an obj-c file to define string identifiers. Or is this the class type?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+     By default it will fall through to the getter for the jsonValue property, but can be different
+     if needed.
+     */
+    var demographicJsonValue: SBBJSONValue? { get }
+}
+
+extension SBAProfileItem {
+    func commonJsonValueGetter() -> SBBJSONValue? {
+        guard let val = self.value else { return NSNull() }
+        switch self.itemType {
```
syoung-smallwisdom, line 85: Yeah, use an enum. :)

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+    
+    open var key: String {
+        get {
+            return sourceDict["key"]! as! String
+        }
+    }
+    
+    open var sourceKey: String {
+        get {
+            return sourceDict["sourceKey"] as! String? ?? self.key
```
syoung-smallwisdom: Shouldn't this be:
````
return sourceDict["sourceKey"] as? String ?? self.key
````

File: BridgeAppSDK/SBAProfileItem.swift [hunk 6]
```diff
@@ -0,0 +1,400 @@
...
+    }
+    
+    open var demographicJsonValue: SBBJSONValue? {
+        get {
+            return self.commonDemographicJsonValue()
+        }
+    }
+}
+
+class SBAKeychainProfileItem: SBAProfileItemBase {
```
syoung-smallwisdom: Is there a reason this isn't `public final` or `open`?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+    private var keychain: SBAKeychainWrapper
```
syoung-smallwisdom: If using a private keychain wrapper, why not just use `let`?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+    
+    public required init(dictionaryRepresentation dictionary: [AnyHashable: Any]) {
+        keychain = SBAUser.shared.keychain
+        
+        let keychainService = dictionary["keychainService"] as! String?
+        let keychainAccessGroup = dictionary["keychainAccessGroup"] as! String?
+        
+        if (keychainService != nil || keychainAccessGroup != nil) {
+            keychain = SBAKeychainWrapper(service: keychainService, accessGroup: keychainAccessGroup)
+        }
```
syoung-smallwisdom:
````
self.keychain = {
    let keychainService = dictionary["keychainService"] as? String
    let keychainAccessGroup = dictionary["keychainAccessGroup"] as? String
    guard (keychainService != nil || keychainAccessGroup != nil) 
    else {
        return SBAUser.shared.keychain
    }
    return SBAKeychainWrapper(service: keychainService, accessGroup: keychainAccessGroup)
}()
````

OR

````
    let keychainService = dictionary["keychainService"] as? String
    let keychainAccessGroup = dictionary["keychainAccessGroup"] as? String
    self.keychain =  (keychainService != nil || keychainAccessGroup != nil) ?
        SBAKeychainWrapper(service: keychainService, accessGroup: keychainAccessGroup) :
        SBAUser.shared.keychain
````

Either implementation will allow you to use `let` rather than `var` which is theoretically more performant and more importantly, will document that this is a property that is readonly and set on init.

File: BridgeAppSDK/SBAProfileItem.swift [hunk 9]
```diff
@@ -0,0 +1,400 @@
...
+        super.init(dictionaryRepresentation: dictionary)
+    }
+    
+    override open var value: Any? {
+        get {
+            var err: NSError?
+            let obj = keychain.object(forKey: key, error: &err)
```
syoung-smallwisdom: Why not use a `guard` here and return `nil`?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+            return self.typedValue(from: obj)
+        }
+        
+        set {
+            do {
+                if newValue == nil {
+                    try keychain.removeObject(forKey: key)
+                } else {
+                    if !self.commonCheckTypeCompatible(newValue: newValue) {
+                        print("Error setting \(key): \(String(describing: newValue)) not compatible with specified type \(itemType)")
```
syoung-smallwisdom: assert?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+                        return
+                    }
+                    guard let secureVal = secureCodingValue(of: newValue) else {
+                        print("Error setting \(key) in keychain: don't know how to convert \(String(describing: newValue))) to NSSecureCoding")
```
syoung-smallwisdom: assert?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+                        return
+                    }
+                    try keychain.setObject(secureVal, forKey: key)
+                }
+            }
+            catch let error as NSError {
```
syoung-smallwisdom: don't need the `as NSError` anymore.

syoung-smallwisdom (commented):
I've gotten about half-way through. Will finish tomorrow.

BridgeAppSDK/SBAProfileDataSource.swift
syoung-smallwisdom: @Erin-Mounts ping

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+    
+    open var profileKey: String {
+        get {
+            return sourceDict["profileKey"]! as! String
+        }
+    }
+    
+    open var sourceKey: String {
+        get {
+            return sourceDict["sourceKey"] as! String? ?? self.profileKey
```
syoung-smallwisdom: Should this be:

````
return sourceDict["sourceKey"] as? String ?? self.profileKey
````

File: BridgeAppSDK/SBAProfileItem.swift [hunk 14]
```diff
@@ -0,0 +1,400 @@
...
+    
+    open var demographicKey: String {
+        get {
+            return sourceDict["demographicKey"] as! String? ?? self.profileKey
+        }
+    }
+    
+    open var itemType: SBAProfileTypeIdentifier {
+        get {
+            return sourceDict["itemType"]! as! SBAProfileTypeIdentifier
```
syoung-smallwisdom: Your itemType is a String whereas SBAProfileTypeIdentifier is a struct. Shouldn't this be:

````
return SBAProfileTypeIdentifier(rawValue: sourceDict["itemType"] as? String ?? "Unknown")
````

Or something like that?  Maybe:

````
guard let rawValue = sourceDict["itemType"] as? String 
else {
    return .string
}
return SBAProfileTypeIdentifier(rawValue: rawValue)
````

File: BridgeAppSDK/SBAProfileItem.swift [hunk 15]
```diff
@@ -0,0 +1,400 @@
...
+    }
+    
+    override open var value: Any? {
+        get {
+            var err: NSError?
+            let obj = keychain.object(forKey: profileKey, error: &err)
+            if let error = err {
+                print("Error accessing keychain \(profileKey): \(error.code) \(error)")
+            }
+            return self.typedValue(from: obj)
```
syoung-smallwisdom, line 265: What happens if `obj` is nil?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+        }
+        
+        set {
+            do {
+                if newValue == nil {
+                    try keychain.removeObject(forKey: profileKey)
+                } else {
+                    if !self.commonCheckTypeCompatible(newValue: newValue) {
+                        print("Error setting \(profileKey): \(String(describing: newValue)) not compatible with specified type \(itemType.rawValue)")
```
syoung-smallwisdom: assert?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+                    }
+                    guard let secureVal = secureCodingValue(of: newValue) else {
+                        print("Error setting \(profileKey) in keychain: don't know how to convert \(String(describing: newValue))) to NSSecureCoding")
+                        return
+                    }
+                    try keychain.setObject(secureVal, forKey: profileKey)
+                }
+            }
+            catch let error as NSError {
+                print("Failed to set \(profileKey): \(error.code) \(error.localizedDescription)")
```
syoung-smallwisdom: assert?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+    
+    public required init(dictionaryRepresentation dictionary: [AnyHashable: Any]) {
+        defaults = SBAUser.shared.bridgeInfo?.userDefaults ?? UserDefaults.standard
+        
+        let userDefaultsSuiteName = dictionary["userDefaultsSuiteName"] as! String?
+        
+        if (userDefaultsSuiteName != nil) {
+            let wasDefaults = defaults
+            defaults = UserDefaults(suiteName: userDefaultsSuiteName) ?? wasDefaults
+        }
```
syoung-smallwisdom: this is kinda confusing. again, I think what you really want is a `let`.

````
private let defaults: UserDefaults

 public required init(dictionaryRepresentation dictionary: [AnyHashable: Any]) {
    defaults = {
        guard let userDefaultsSuiteName = dictionary["userDefaultsSuiteName"] as? String,
            let userDefaults = UserDefaults(suiteName: userDefaultsSuiteName) 
        else {
            return SBAUser.shared.bridgeInfo?.userDefaults ?? UserDefaults.standard
        }
        return userDefaults
    }()
        super.init(dictionaryRepresentation: dictionary)
}
````

File: BridgeAppSDK/SBAProfileItem.swift [hunk 19]
```diff
@@ -0,0 +1,400 @@
...
+        get {
+            return typedValue(from: defaults.object(forKey: profileKey) as? PlistValue)
+        }
+        
+        set {
+            if newValue == nil {
+                defaults.removeObject(forKey: profileKey)
+            } else {
+                if !self.commonCheckTypeCompatible(newValue: newValue) {
+                    print("Error setting \(profileKey): \(String(describing: newValue)) not compatible with specified type\(itemType.rawValue)")
```
syoung-smallwisdom: assert?

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: nope

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -0,0 +1,400 @@
...
+                    return
+                }
+                guard let plistVal = pListValue(of: newValue) else {
+                    print("Error setting \(profileKey) in user defaults: don't know how to convert \(String(describing: newValue)) to PlistValue")
```
syoung-smallwisdom: assert?

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: because it would only let me set it once in the initializer--fixed by using a temp var until I'm sure what to set it to

BridgeAppSDK/SBAProfileManager.swift
```diff
@@ -0,0 +1,239 @@
...
+     @param key The profileKey of the item whose value is to be set.
+     */
+    func setValue(_ value: Any?, forProfileKey key: String) throws
+    
+    /**
+     Set up and return a view controller for displaying a Profile view.
+     
+     @return A view controller for displaying the Profile view.
+     */
+    func profileViewController() -> UIViewController?
```
syoung-smallwisdom: Set the availability to iOS or rethink this; mixing UI with model.

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: whoops, missed that when updating

BridgeAppSDK/SBAProfileManager.swift
```diff
@@ -0,0 +1,239 @@
...
+
+
+open class SBAProfileManager: SBADataObject, SBAProfileManagerProtocol, SBAProfileDataSource {
+
+    /**
+     The shared instance of the profile manager. Loads from SBAProfileItemsJSONFilename, which
+     defaults to "ProfileItems".
+     */
+    static let shared: SBAProfileManagerProtocol? = {
+        guard let json = SBAResourceFinder.shared.json(forResource: SBAProfileItemsJSONFilename),
```
syoung-smallwisdom: Since you have this setup as an open class, do you want to check the json for a "classType" first or else add a comment that the app would need to add an override into the ClassTypeMap plist.

BridgeAppSDK/SBAProfileManager.swift
```diff
@@ -0,0 +1,239 @@
...
+                let sharedProfileManager = SBAClassTypeMap.shared.object(with:json, classType:SBAProfileManagerClassType) as? SBAProfileManagerProtocol
+            else { return nil }
+        return sharedProfileManager
+    }()
+
+    private dynamic var items: [SBAProfileItem] = []
```
syoung-smallwisdom, line 117: Didn't know that a property could be private and still work with KVO. huh. ;)

BridgeAppSDK/SBAProfileManager.swift
```diff
@@ -0,0 +1,239 @@
...
+    lazy private var itemsKeys: [String] = {
+        var allKeys: [String] = []
+        for item in self.items {
+            allKeys.append(item.profileKey)
+        }
+        return allKeys
```
syoung-smallwisdom:
````
return self.items.map({ $0.profileKey })
````

File: BridgeAppSDK/SBAProfileManager.swift [hunk 25]
```diff
@@ -0,0 +1,239 @@
...
+    }()
+    
+    lazy private var itemsMap: [String: SBAProfileItem] = {
+        var allItems: [String: SBAProfileItem] = [:]
+        for item in self.items {
+            allItems[item.profileKey] = item
+        }
+        return allItems
```
syoung-smallwisdom, line 127: Wow, this really feels like something that needs an extension on Dictionary. ;)

BridgeAppSDK/SBAProfileManager.swift
```diff
@@ -0,0 +1,239 @@
...
+   
+    // MARK: SBADataObject overrides
+    
+    override open func dictionaryRepresentationKeys() -> [String] {
+        return super.dictionaryRepresentationKeys().appending(#keyPath(items))
+    }
+    
+    override open func defaultValue(forKey key: String) -> Any? {
+        if key == "items" {
+            return [] as [String]
```
syoung-smallwisdom: This is kind of confusing. Is this what you are trying to do?

````
return [SBAProfileItem]()
````

File: BridgeAppSDK/SBAProfileItem.swift (hunk 5 above)
Comment (Erin-Mounts) [reply BridgeAppSDK/SBAProfileItem.swift]: indeed it should--Xcode's Swift suggestions when I get it wrong seem to be wrong as often as right :/

File: BridgeAppSDK/SBAProfileSection.swift [hunk 27]
```diff
@@ -0,0 +1,58 @@
...
+public protocol SBAProfileSection: NSObjectProtocol {
+    var title: String? { get }
+    var items: [SBAProfileItem] { get }
+}
+
+public protocol SBAProfileTableItem: NSObjectProtocol {
+    var profileItemKey: String? { get }
+    var title: String { get }
+    var isEditable: Bool { get }
+    var onSelectedViewController: UIViewController? { get }
```
syoung-smallwisdom, line 46: Again, mixing UI specific to iOS with model. Also, I can see a need later for a UI that includes inline editing. It's gonna happen. ¯\_(ツ)_/¯

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: because someone might want to subclass this and have typedValue() return e.g. NSNull() or some default value if it's nil.

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: yep thanks

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: 👍

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: indeed

syoung-smallwisdom (approved):
Syntax nits. Otherwise, looks good.

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: This implementation of typedValue will just return nil. Subclasses might override to do something else.

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: yep

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: 👍

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: fixed an entirely different way

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: sure

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: ayup

BridgeAppSDK/SBAProfileManager.swift
Erin-Mounts: thank you, couldn't remember the map syntax

BridgeAppSDK/SBAProfileManager.swift
Erin-Mounts: not sure how you mean?

BridgeAppSDK/SBAProfileManager.swift
Erin-Mounts: oh yeah derp, not sure what I was thinking...

BridgeAppSDK/SBAProfileManager.swift
Erin-Mounts: good point (assuming I'm understanding your point)...

BridgeAppSDK/SBAProfileManager.swift
syoung-smallwisdom: Just that mapping an array to a dictionary is something that I seem to do a lot.

PR #216, opened 2017-05-11
Add to the schedule filtering and allow schedule manager to load date range.

syoung-smallwisdom:
Lots of rope.


Erin-Mounts (approved):
looks good

PR #217, opened 2017-05-11
bare bones demo of profile table items

Erin-Mounts:
with some useful and necessary profile item/manager updates

### DIFF
--- a/BridgeAppSDK/SBAProfileDataSource.swift
+++ b/BridgeAppSDK/SBAProfileDataSource.swift
@@ -88,10 +88,13 @@ open class SBAProfileDataSourceObject: SBADataObject, SBAProfileDataSource {
      before accessing this, and map *that* in your ClassTypeMap.plist; or just set SBAProfileSectionClassType
      to the fully qualified class name without setting up any mappings.
      */
-    public static let shared: SBAProfileDataSource? = {
+    public static let shared: SBAProfileDataSource = {
         guard let json = SBAResourceFinder.shared.json(forResource: SBAProfileJSONFilename),
             let sharedProfileDataSource = SBAClassTypeMap.shared.object(with:json, classType:SBAProfileDataSourceClassType) as? SBAProfileDataSource
-            else { return nil }
+            else {
+                assertionFailure("Couldn't find the shared profile data source.")
+                return SBAProfileDataSourceObject(dictionaryRepresentation: [:])
+        }
         return sharedProfileDataSource
     }()
     
@@ -118,7 +121,7 @@ open class SBAProfileDataSourceObject: SBADataObject, SBAProfileDataSource {
     }
     
     open func numberOfRows(for section: Int) -> Int {
-        if section >= sections.count { return 0 } // out of range
+        guard section < sections.count else { return 0 } // out of range
         return sections[section].items.count
     }
     
@@ -126,14 +129,13 @@ open class SBAProfileDataSourceObject: SBADataObject, SBAProfileDataSource {
         let section = indexPath.section
         let row = indexPath.row
         
-        if section >= sections.count { return nil }
-        if row >= sections[section].items.count { return nil }
+        guard (section < sections.count) && (row < sections[section].items.count) else { return nil }
         
         return sections[section].items[row]
     }
     
     open func title(for section: Int) -> String? {
-        if section >= sections.count { return nil }
+        guard section < sections.count else { return nil }
         return sections[section].title
     }
 }
--- a/BridgeAppSDK/SBAProfileDataSource.swift
+++ b/BridgeAppSDK/SBAProfileDataSource.swift
@@ -33,6 +33,9 @@
 
 import Foundation
 
+public var SBAProfileJSONFilename = "Profile"
+public var SBAProfileDataSourceClassType = "ProfileDataSource"
+
 @objc
 public protocol SBAProfileDataSource: class {
     /**
@@ -50,11 +53,11 @@ public protocol SBAProfileDataSource: class {
     func numberOfRows(for section: Int) -> Int
     
     /**
-     The profile item at the given index.
-     @param indexPath   The index path for the profile item.
+     The profile table item at the given index.
+     @param indexPath   The index path for the profile table item.
      */
-    @objc(profileItemAtIndexPath:)
-    func profileItem(at indexPath: IndexPath) -> SBAProfileItem?
+    @objc(profileTableItemAtIndexPath:)
+    func profileTableItem(at indexPath: IndexPath) -> SBAProfileTableItem?
     
     /**
      Called when a row is selected.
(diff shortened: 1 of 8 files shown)


BridgeAppSDK/SBAProfileDataSource.swift
```diff
@@ -33,6 +33,9 @@
 
 import Foundation
 
+public var SBAProfileJSONFilename = "Profile"
+public var SBAProfileDataSourceClassType = "ProfileDataSource"
```
syoung-smallwisdom, line 37: Why `var` rather than `let`? I'm not actually sure how public vars that aren't in a class work.

BridgeAppSDK/SBAProfileDataSource.swift
```diff
@@ -71,3 +74,66 @@ public protocol SBAProfileDataSource: class {
...
+    /**
+     The shared instance of the profile data source. Loads from SBAProfileJSONFilename, which
+     defaults to "Profile".
+     
+     If you subclass SBAProfileDataSourceObject, or just want this method to return a different class that
+     implements the SBAProfileDataSource protocol, you will need to override the mapping for the default
+     value of SBAProfileDataSourceClassType ("ProfileDataSource") in your ClassTypeMap.plist to map to your
+     class instead of this one. Alternatively you could change the value of SBAProfileSectionClassType
+     before accessing this, and map *that* in your ClassTypeMap.plist; or just set SBAProfileSectionClassType
+     to the fully qualified class name without setting up any mappings.
```
syoung-smallwisdom, line 89: thank you. :)

BridgeAppSDK/SBAProfileDataSource.swift
```diff
@@ -71,3 +74,66 @@ public protocol SBAProfileDataSource: class {
...
+        }
+    }
+
+    // MARK: SBAProfileDataSource
+    
+    open func numberOfSections() -> Int {
+        return sections.count
+    }
+    
+    open func numberOfRows(for section: Int) -> Int {
```
syoung-smallwisdom, line 123: `@objc` ?

BridgeAppSDK/SBAProfileDataSource.swift
```diff
@@ -71,3 +74,66 @@ public protocol SBAProfileDataSource: class {
...
+        if section >= sections.count { return 0 } // out of range
+        return sections[section].items.count
+    }
+    
+    open func profileTableItem(at indexPath: IndexPath) -> SBAProfileTableItem? {
+        let section = indexPath.section
+        let row = indexPath.row
+        
+        if section >= sections.count { return nil }
+        if row >= sections[section].items.count { return nil }
```
syoung-smallwisdom:
````
guard (section < sections.count) && (row < sections[section].items.count) else { return nil }
````

File: BridgeAppSDK/SBAProfileDataSource.swift [hunk 5]
```diff
@@ -71,3 +74,66 @@ public protocol SBAProfileDataSource: class {
...
+    }
+
+    // MARK: SBAProfileDataSource
+    
+    open func numberOfSections() -> Int {
+        return sections.count
+    }
+    
+    open func numberOfRows(for section: Int) -> Int {
+        if section >= sections.count { return 0 } // out of range
```
syoung-smallwisdom:
````
guard section < sections.count else { return 0 }
````

File: BridgeAppSDK/SBAProfileDataSource.swift [hunk 6]
```diff
@@ -71,3 +74,66 @@ public protocol SBAProfileDataSource: class {
...
+        let row = indexPath.row
+        
+        if section >= sections.count { return nil }
+        if row >= sections[section].items.count { return nil }
+        
+        return sections[section].items[row]
+    }
+    
+    open func title(for section: Int) -> String? {
+        if section >= sections.count { return nil }
```
syoung-smallwisdom:
````
guard section < sections.count else { return nil }
````

File: BridgeAppSDK/SBAProfileSection.swift [hunk 7]
```diff
@@ -34,25 +34,117 @@
...
     }
+
+    override open func defaultValue(forKey key: String) -> Any? {
+        if key == #keyPath(items) {
+            return [SBAProfileTableItem]()
+        } else {
+            return super.defaultValue(forKey: key)
+        }
+    }
+}
```
syoung-smallwisdom: Complete and total nit: Could you please add a whitespace line?

BridgeAppSDK/SBAProfileSection.swift
```diff
@@ -34,25 +34,117 @@
...
+            let key = #keyPath(detail)
+            return sourceDict[key] as? String ?? ""
+        }
+    }
+    
+    open var isEditable: Bool {
+        get {
+            let key = #keyPath(isEditable)
+            guard let editable = sourceDict[key] as? Bool else { return false }
+            return editable
```
syoung-smallwisdom:
````
return sourceDict[key] as? Bool ?? false
````

Is more consistent with the other properties.

File: BridgeAppSDK/SBAProfileDataSource.swift [hunk 9]
```diff
@@ -71,3 +74,66 @@ public protocol SBAProfileDataSource: class {
...
+     defaults to "Profile".
+     
+     If you subclass SBAProfileDataSourceObject, or just want this method to return a different class that
+     implements the SBAProfileDataSource protocol, you will need to override the mapping for the default
+     value of SBAProfileDataSourceClassType ("ProfileDataSource") in your ClassTypeMap.plist to map to your
+     class instead of this one. Alternatively you could change the value of SBAProfileSectionClassType
+     before accessing this, and map *that* in your ClassTypeMap.plist; or just set SBAProfileSectionClassType
+     to the fully qualified class name without setting up any mappings.
+     */
+    public static let shared: SBAProfileDataSource? = {
```
syoung-smallwisdom: It's confusing to declare this as optional and then force it when you use the lazy initializer below without any explanation for why that is throwing an error.  Instead of returning nil can you do this:

````
else {
    assertionFailure("Couldn't find the shared profile data source.")
    return SBAProfileDataSourceObject(dictionary: [:])
}
````

Or if it makes more sense, just return an empty source without an assert and then have the lazy initializer throw an assert if the number of sections == 0.

File: BridgeAppSDK/SBAProfileManager.swift [hunk 10]
```diff
@@ -107,7 +106,7 @@ open class SBAProfileManager: SBADataObject, SBAProfileManagerProtocol, SBAProfi
      before accessing this, and map *that* in your ClassTypeMap.plist; or just set SBAProfileManagerClassType
      to the fully qualified class name without setting up any mappings.
      */
-    static let shared: SBAProfileManagerProtocol? = {
+    public static let shared: SBAProfileManagerProtocol? = {
```
syoung-smallwisdom, line 109: Again, same issue with using an optional.

syoung-smallwisdom (approved):
A number of syntax nits. Otherwise, looks good. Merge when ready.

BridgeAppSDK/SBAProfileDataSource.swift
Erin-Mounts: ...was assuming they'd essentially act like globals; making them var allows folks to change them before they get used so they load from differently-named files without modifying any code--not a huge gain, could switch to let if you prefer

syoung-smallwisdom: Not concerned, just curious.

BridgeAppSDK/SBAProfileDataSource.swift
Erin-Mounts: do I need to do it here too even though it's defined in the protocol?

BridgeAppSDK/SBAProfileSection.swift
Erin-Mounts: whoops

BridgeAppSDK/SBAProfileDataSource.swift
Erin-Mounts: thanks

BridgeAppSDK/SBAProfileDataSource.swift
Erin-Mounts: thanks

BridgeAppSDK/SBAProfileDataSource.swift
Erin-Mounts: thanks

BridgeAppSDK/SBAProfileDataSource.swift
syoung-smallwisdom: Not sure. I guess it doesn't really make sense that you would need to .

PR #219, opened 2017-05-14
allow SBAUser.shared as a profile item source

Erin-Mounts:


BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -408,3 +408,28 @@ open class SBAUserDefaultsProfileItem: SBAProfileItemBase {
         return retVal
     }
 }
+
+open class SBAUserProfileItem: SBAProfileItemBase {
+    open static var user: SBAUser = {
```
syoung-smallwisdom: Why not a non-static internal lazy var that you can use for testing?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -408,3 +408,28 @@ open class SBAUserDefaultsProfileItem: SBAProfileItemBase {
...
+        return SBAUser.shared
+    }()
+    
+    override open var value: Any? {
+        get {
+            // need to special-case accessing properties that are defined in protocol extensions
+            // because they're not available via KVC
+            switch sourceKey {
+            case "givenName":
```
syoung-smallwisdom: String enum?

BridgeAppSDK/SBAProfileItem.swift
syoung-smallwisdom: Also, your tests should use MockUser.

syoung-smallwisdom (approved):
Please change the shared user to an instance. That allows testing with MockUser. Otherwise, looks good.

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: good idea

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: This turns out to be a lot less trivial than it looks at first glance, since SBAUserProfileItem depends on user implementing SBANameDataSource. MockUser implements SBAUserWrapper but not SBANameDataSource, and even if it did, since it's defined in ObjC it wouldn't have access to the givenName and fullName methods which are defined only in a protocol extension.

Having the tests run against the real SBAUser shared instance will let us know if any changes to that object would break this code, and I think it's stable enough at this point that it doesn't necessarily make these tests any more fragile than they would be with a mock object.

syoung-smallwisdom (commented):
looks good

PR #221, opened 2017-05-16
Modify the profile manager and SBAUser to have one-way pointers

syoung-smallwisdom:
These changes allow accessing items that are stored and/or displayed by the profile manager while using the same method to access those items via the SBAUser (if defined for the profile manager).

### DIFF
--- a/BridgeAppSDK/SBAProfileItem.swift
+++ b/BridgeAppSDK/SBAProfileItem.swift
@@ -77,6 +77,11 @@ public protocol SBAProfileItem: NSObjectProtocol {
      if needed.
      */
     var demographicJsonValue: SBBJSONValue? { get }
+    
+    /**
+     Is the value read-only?
+     */
+    var readonly: Bool { get }
 }
 
 extension SBAProfileItem {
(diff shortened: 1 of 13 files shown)


BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -239,54 +244,60 @@ open class SBAProfileItemBase: NSObject, SBAProfileItem {
...
         
         set {
             do {
                 if newValue == nil {
-                    try keychain.removeObject(forKey: profileKey)
+                    try keychain.removeObject(forKey: sourceKey)
                 } else {
                     if !self.commonCheckTypeCompatible(newValue: newValue) {
-                        assert(false, "Error setting \(profileKey): \(String(describing: newValue)) not compatible with specified type \(itemType.rawValue)")
+                        assert(false, "Error setting \(sourceKey): \(String(describing: newValue)) not compatible with specified type \(itemType.rawValue)")
```
Erin-Mounts: I'm thinking this particular message maybe should refer to both the profileKey and the sourceKey to avoid confusion about what it means.

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -350,40 +370,39 @@ extension Data: PlistValue {}
...
         }
         
         set {
             if newValue == nil {
-                defaults.removeObject(forKey: profileKey)
+                defaults.removeObject(forKey: sourceKey)
             } else {
                 if !self.commonCheckTypeCompatible(newValue: newValue) {
-                    assert(false, "Error setting \(profileKey): \(String(describing: newValue)) not compatible with specified type\(itemType.rawValue)")
+                    assert(false, "Error setting \(sourceKey): \(String(describing: newValue)) not compatible with specified type\(itemType.rawValue)")
```
Erin-Mounts: refer to both the profileKey and the sourceKey?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -409,27 +428,43 @@ open class SBAUserDefaultsProfileItem: SBAProfileItemBase {
...
-    }()
+open class SBAUserProfileItem: SBAKeychainProfileItem, SBANameDataSource {
     
     override open var value: Any? {
+        
         get {
-            // need to special-case accessing properties that are defined in protocol extensions
-            // because they're not available via KVC
-            switch SBAUserProtocolExtension(sourceKey) {
-            case SBAUserProtocolExtension.givenName:
```
Erin-Mounts, line 422: why remove givenName as an option here?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -409,27 +428,43 @@ open class SBAUserDefaultsProfileItem: SBAProfileItemBase {
...
         set {
-            user.setValue(newValue, forKey: sourceKey)
+            if !readonly {
+                super.value = newValue
+            }
         }
     }
+    
+    override open var readonly: Bool {
+        return self.sourceKey == SBAProfileSourceKey.fullName.rawValue
```
Erin-Mounts, line 459: if givenName were supported it would need to be included here too

BridgeAppSDKTests/SBAProfileManagerTests.swift
```diff
@@ -26,31 +29,33 @@ class SBAProfileManagerTests: ResourceTestCase {
     }
     
     func testProfileKeys() {
-        guard let keys = ProfileManager?.profileKeys() else {
+        guard let keys = profileManager?.profileKeys() else {
             XCTFail("No ProfileManager instance")
             return
         }
-        XCTAssert(keys.count == 6, "expected 6 keys, got \(keys.count)")
+        XCTAssert(keys.count == 9, "expected 6 keys, got \(keys.count)")
```
Erin-Mounts: change it in the string too

BridgeAppSDKTests/SBAProfileManagerTests.swift
```diff
@@ -26,31 +29,33 @@ class SBAProfileManagerTests: ResourceTestCase {
...
     }
     
     func testProfileItems() {
-        guard let items = ProfileManager?.profileItems() else {
+        guard let items = profileManager?.profileItems() else {
             XCTFail("No ProfileManager instance")
             return
         }
-        XCTAssert(items.count == 6, "expected 6 items, got \(items.count)")
+        XCTAssert(items.count == 9, "expected 6 items, got \(items.count)")
```
Erin-Mounts: 6 --> 9 in the string too

Erin-Mounts (approved):
Some nits, otherwise looks good

BridgeAppSDK/SBAProfileItem.swift
syoung-smallwisdom: given name is the default value that maps to "name" for the case where the first name field is used to describe the full name and the last name field is unused. I'm getting around the loop of this thing points to the user and the user points back at this (which results in infinite loops of wacky madness and is why I'm not a big fan of singletons) by *only* having the user drop through to the shared profile manager.

`SBAUser -> ProfileManager -> ProfileItems`

These items access the given/family name directly. This still requires a few keys that have to be defined in the json and match the hardcoded values *but* it allows for a flat list without circular references.

BridgeAppSDK/SBAProfileItem.swift
syoung-smallwisdom: *givenName* is read/write.  *preferredName* is read/write but if nil, return given name.

PR #222, opened 2017-05-17
Fix drop-through for setting values to keychain

syoung-smallwisdom:
And updates for registration UI/UX, profile keys and email verification.


Erin-Mounts (approved):
Looks good

PR #224, opened 2017-05-18
Refactor profile items

syoung-smallwisdom:
Accidentally closed

### DIFF
--- a/BridgeAppSDK/SBAProfileItem.swift
+++ b/BridgeAppSDK/SBAProfileItem.swift
@@ -82,6 +82,12 @@ public protocol SBAProfileItem: NSObjectProtocol {
      Is the value read-only?
      */
     var readonly: Bool { get }
+    
+    /**
+     Some of the stored value types have a unit associated with them that is used to 
+     build the model object into an `HKQuantity`.
+     */
+    var unit: HKUnit? { get }
 }
 
 extension SBAProfileItem {
@@ -104,7 +110,8 @@ extension SBAProfileItem {
             return val as? NSNumber
             
         case SBAProfileTypeIdentifier.hkQuantity:
-            return val as? NSNumber
+            guard let quantity = val as? HKQuantity else { return nil }
+            return NSNumber(value: quantity.doubleValue(for: self.unit ?? commonDefaultUnit()))
             
         default:
             return nil
@@ -119,8 +126,7 @@ extension SBAProfileItem {
         
         switch self.itemType {
         case SBAProfileTypeIdentifier.string:
-            guard let val = value as? String else { return }
-            self.value = val
+            self.value = String(describing: value)
             
         case SBAProfileTypeIdentifier.number:
             guard let val = value as? NSNumber else { return }
@@ -142,7 +148,7 @@ extension SBAProfileItem {
             
         case SBAProfileTypeIdentifier.hkQuantity:
             guard let val = value as? NSNumber else { return }
-            self.value = val
+            self.value = HKQuantity(unit: self.unit ?? commonDefaultUnit(), doubleValue: val.doubleValue)
             
         default:
             break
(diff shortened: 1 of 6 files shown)


BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -163,65 +169,103 @@ extension SBAProfileItem {
...
             return newValue as? HKBiologicalSex != nil
             
         case SBAProfileTypeIdentifier.hkQuantity:
-            return newValue as? NSNumber != nil
+            guard let quantity = newValue as? HKQuantity else { return false }
+            return quantity.is(compatibleWith: self.unit ?? commonDefaultUnit())
             
         default:
-            return false
+            return true   // Any extended type isn't included in the common validation
```
Erin-Mounts, line 197: Good point, but this way any value is accepted for any extended type without any kind of validation. Currently we have to add support for any new types here, and then they're supported by all SBAProfileItem implementations. To be honest, I couldn't think of a clean way to extend SBAProfileItem to support new types outside of BridgeAppSDK. The best I could come up with was, if we want to allow that, for each commonXxx() func here we'd make an open xxx() func in SBAProfileItemBase that calls the corresponding commonXxx() and have the classes use those methods instead of the commonXxx() ones directly. Then if you wanted to support an extended type without adding it here, you'd have to subclass whatever profile item class(es) need to support it and override those methods, in which case returning false from super.checkTypeCompatible() would let you know it wasn't a type supported by the parent class so you should do your own checking etc.; and then you'd use your subclasses wherever you needed to support the extended types. If this returns true for unknown types then this approach doesn't work; on the other hand it didn't feel like a particularly elegant or clean approach so I didn't go ahead and build it that way from the start. I couldn't come up with a way to do it via further extensions. Maybe you have a better idea. Anyway I'm OK with this change if we need it for something we're doing to work, just wanted to clarify why I did it the other way originally.

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -163,65 +169,103 @@ extension SBAProfileItem {
...
      The value property is used to get and set the profile item's value in whatever internal data
      storage is used by the implementing class.
      */
-    open var value: Any?
+    open var value: Any? {
+        get {
+            // Look at the sourceKey, if not found then fall back to the fallback key and check that
+            let value = storedValue(forKey: sourceKey)
+            if value == nil, let fallback = fallbackKey {
+                return storedValue(forKey: fallback)
```
Erin-Mounts, line 230: I had had in mind the the fallbackKey would be used as a profileKey via the profile manager so it could fall back to a different type of storage if need be, but if this does what we need, tbh it's safer since it doesn't allow for accidentally creating "fallback loops". 👍

BridgeAppSDKTests/SBAProfileManagerTests.swift
```diff
@@ -60,14 +60,14 @@ class SBAProfileManagerTests: ResourceTestCase {
             XCTAssert(fullNameItem!.sourceKey == "fullName", "expected fullNameItem.sourceKey to be fullName, but it's \(fullNameItem!.sourceKey)")
             XCTAssert(fullNameItem!.demographicKey == fullNameItem!.profileKey, "expected fullNameItem.demographicKey to be \(fullNameItem!.profileKey), but it's \(fullNameItem!.demographicKey)")
             XCTAssert(fullNameItem!.itemType == .string, "expected fullNameItem.itemType to be String, but it's \(fullNameItem!.itemType.rawValue)")
-            let typedItem = fullNameItem as? BridgeAppSDK.SBAUserProfileItem
+            let typedItem = fullNameItem as? BridgeAppSDK.SBAKeychainProfileItem
```
Erin-Mounts: SBAFullNameProfileItem

BridgeAppSDKTests/SBAProfileManagerTests.swift
```diff
@@ -60,14 +60,14 @@ class SBAProfileManagerTests: ResourceTestCase {
...
             XCTAssertNotNil(typedItem, "fullNameItem is not an SBAUserProfileItem: \(String(describing: fullNameItem))")
```
Erin-Mounts: SBAUserProfileItem --> SBAFullNameProfileItem

BridgeAppSDKTests/SBAProfileManagerTests.swift
```diff
@@ -60,14 +60,14 @@ class SBAProfileManagerTests: ResourceTestCase {
...
         }
         if externalIdItem != nil {
             XCTAssert(externalIdItem!.sourceKey == "externalId", "expected externalIdItem.sourceKey to be externalId, but it's \(externalIdItem!.sourceKey)")
             XCTAssert(externalIdItem!.demographicKey == externalIdItem!.profileKey, "expected externalIdItem.demographicKey to be \(externalIdItem!.profileKey), but it's \(externalIdItem!.demographicKey)")
             XCTAssert(externalIdItem!.itemType == .string, "expected externalIdItem.itemType to be String, but it's \(externalIdItem!.itemType.rawValue)")
-            let typedItem = externalIdItem as? BridgeAppSDK.SBAUserProfileItem
+            let typedItem = externalIdItem as? BridgeAppSDK.SBAKeychainProfileItem
             XCTAssertNotNil(typedItem, "externalIdItem is not an SBAUserProfileItem: \(String(describing: externalIdItem))")
```
Erin-Mounts: SBAUserProfileItem --> SBAKeychainProfileItem

Erin-Mounts (approved):
couple of test nits, and a couple of food-for-thought type comments, otherwise looks good, merge when ready

BridgeAppSDK/SBAProfileItem.swift
syoung-smallwisdom: This came up b/c Eric modeled the externalId to return a number-type question result. I ended up removing that value from the profile items, but it did occur to me that the server handles strings this way so it seemed as if there was precedent.

BridgeAppSDK/SBAProfileItem.swift
syoung-smallwisdom: yup, that's why I went with this. there are enough loops already b/c of the SBAUser. Also, I kind of figured that it doesn't totally make sense for the question type wouldn't change (and therefore where it is stored wouldn't change). You can still override the default implementation for either `storedValue()` or `value`.

PR #225, opened 2017-05-18
Add Array and Dictionary element types.

syoung-smallwisdom:
Next chunk of work required to link the Demographics survey to the profile. "race" question allows multiple answers.

Note: this does not include mapping the values -> localized text.


Erin-Mounts (approved):
looks good

PR #226, opened 2017-05-19
Localize profile item display text

syoung-smallwisdom:
Handle localization and formating for the case where the profile item is a list of choices.

This will handle the case where the selected value maps to a list of choices defined in the localized ProfileQuestions json file.


Erin-Mounts (approved):
looks good

PR #227, opened 2017-05-21
bug/cancel_withdraw

dephillipsmichael:
fixed bug where user can get stuck on withdraw screen and they want to cancel


syoung-smallwisdom (approved):
Whoops. Thanks.

PR #228, opened 2017-05-22
Onboarding login

syoung-smallwisdom:
Split out external ID usage into (a) set the external ID for a registered user, (b) store the external ID for registration if not registered, OR (c) login via external ID. Also fixed onboarding manager to return appropriate steps for login verse signup.


Erin-Mounts (approved):
looks good

PR #231, opened 2017-05-24
Stub out a view controller that will push the brain baseline vc when rotated

syoung-smallwisdom:
Because of the framework not being included at this level this is just stubbed out, but I'm including here for reuse.


BridgeAppSDK/SBABrainBaselineStepViewController.swift
```diff
@@ -0,0 +1,229 @@
...
+
+        // if user has consented we will have their birthdate (at least the year).
+        let year = Calendar.current.component(.year, from: self.birthdate!)
+        
+        let gender = SBAProfileManager.shared?.value(forProfileKey: SBAProfileInfoOption.gender.rawValue)
+        let genderString: String? = (gender as? String) ?? ((gender as? HKBiologicalSex)?.demographicDataValue as String?)
+        
+        // TODO: emm2017-04-30 Set other demographic values we want to compare against, e.g. education
+        
+        return ["birth_year" : year,
```
Erin-Mounts: the list of demographic attributes the data can be grouped by will be different for different studies, so maybe in a protocol extension isn't ideal for this?

BridgeAppSDK/SBABrainBaselineStepViewController.swift
```diff
@@ -0,0 +1,229 @@
...
+                self.setStoredAnswer(uuid, forKey: uniqueIdentifierUserKey)
+                return uuid
+        }
+        
+        // If the md5 was created then return that
+        return md5
+    }
+    
+    /**
+     Mapping of key/value pairs for
```
Erin-Mounts: for...?

Erin-Mounts (approved):
one nit, one question, otherwise looks good, merge when ready

PR #232, opened 2017-05-25
IA-122 check that started on and finished on before expired on

syoung-smallwisdom:


syoung-smallwisdom:
Need to handle this in elevateMS.

PR #233, opened 2017-05-25
Caching Bug

dephillipsmichael:
Switched to using cache first for activities so that client data will save.

Should this be on master?


syoung-smallwisdom (approved):
looks good.

This is already work-around code for the limitations of the server. Should be fine for now since elevateMS is the only study using it.

PR #237, opened 2017-05-30
reverse_cache_fix

dephillipsmichael:
Revert "Switched to using cache first for activities so that client data will save"

This reverts commit 9095f83e52c820f96bc2877cecc855836b6fa3f1.


syoung-smallwisdom (approved):
looks good

PR #242, opened 2017-05-31
bug/recording_incorrect_activities

dephillipsmichael:
Added an additional check to make sure we have the correct schedule for the taskViewController.
With this in, everything works correctly when completing symptoms, factors, or activities.


syoung-smallwisdom (changes requested):
Please use an override in elevateMS and create a JIRA for fixing here. This isn't going to work as a suitable logic for all apps and schedules. Also, expiresOn can be null so this will crash bc of the forced unwrap.

dephillipsmichael:
@syoung-smallwisdom do you mind creating the Jira?  I'm not sure I would be able to provide the relevant details.  I'm not as familiar with all the other apps and schedules.

dephillipsmichael:
Closing because I will open a different PR in elevate MS

PR #244, opened 2017-06-03
Update user info from auth delegate callback

Erin-Mounts:
requires https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/155

### DIFF
--- a/BridgeAppSDK/SBAUserWrapper+Bridge.swift
+++ b/BridgeAppSDK/SBAUserWrapper+Bridge.swift
@@ -518,9 +518,9 @@ public extension SBAUserWrapper {
     }
     
     /**
-     Call this any time the user session info gets updated.
+     Call this any time the user session info gets updated elsewhere.
      */
-    public func updateFromUserSessionInfo(_ info: SBBUserSessionInfo) {
+    func updateFromUserSessionInfo(_ info: SBBUserSessionInfo) {
         self.updateFromUserSession(info)
     }
     
--- a/BridgeAppSDK/SBAUserWrapper+Bridge.swift
+++ b/BridgeAppSDK/SBAUserWrapper+Bridge.swift
@@ -517,7 +517,14 @@ public extension SBAUserWrapper {
         }
     }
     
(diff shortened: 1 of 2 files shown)


BridgeAppSDK/SBAUserWrapper+Bridge.swift
```diff
@@ -517,7 +517,14 @@ public extension SBAUserWrapper {
...
     
-    fileprivate func updateFromUserSession(_ response: SBAUserSessionInfoWrapper) {
+    /**
+     Call this any time the user session info gets updated.
+     */
+    public func updateFromUserSessionInfo(_ info: SBBUserSessionInfo) {
+        self.updateFromUserSession(info)
+    }
+    
+    private func updateFromUserSession(_ response: SBAUserSessionInfoWrapper) {
```
syoung-smallwisdom: not sure I understand the value of having both public and private instead of just making the wrapper one public.

syoung-smallwisdom (approved):
One question. Otherwise, looks good. Merge when ready.

BridgeAppSDK/SBAUserWrapper+Bridge.swift
Erin-Mounts: Did this because SBAUserSessionInfoWrapper is an internal protocol. Would it make more sense to make that protocol public?

syoung-smallwisdom: Ah, right. Why do either need to be public? What's the reason for an app to call this method?

Erin-Mounts: well it technically doesn't need to be public I suppose, but at least not fileprivate so the SBBAuthManagerDelegate instance (SBAUser) can call it when it gets a delegate callback from BridgeSDK.

syoung-smallwisdom: sure, I only made it fileprivate b/c the only implementation (thus far) was in the same file so it made sense. ;)

PR #245, opened 2017-06-03
support for profile navigation

Erin-Mounts:
add a field to table items to indicate what should happen when the item is selected


syoung-smallwisdom (changes requested):
Not sure I understand the intension? Does this map to a selector (risky)? Or enum? If it's an enum, use an extensible enum instead. (See `SBASurveyItemType` or `SBAOnboardingSectionType`)

Erin-Mounts:
Now using an extensible string enum instead of a raw string.

syoung-smallwisdom (approved):
looks good. merge when ready.

PR #246, opened 2017-06-07
Custom steps

syoung-smallwisdom:
Some overlap with the stuff that @joshbruhin is doing with elevateMS onboarding. This is focused on migrating towards a step presentation that can use the custom steps where possible but will fall back to using those defined by ResearchKit where a custom step is not defined.


BridgeAppSDK/SBARoundedButton.swift
```diff
@@ -53,14 +53,18 @@ import UIKit
     
     override open var isEnabled: Bool {
         didSet {
-            // simple way to show disabled state
+            // show as disabled by lowering opacity unless alpha is used to set hidden
```
Erin-Mounts, line 56: where is alpha used to set hidden? might be worth mentioning here for future reference

BridgeAppSDK/SBARoundedButton.swift
```diff
@@ -85,6 +117,11 @@ import UIKit
     }
     
     func refreshView() {
+        guard self.alpha > 0.9 else {
```
Erin-Mounts, line 120: would it make sense for 0.9 be a constant instead?

BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -361,6 +361,18 @@ open class SBABaseScheduledActivityManager: NSObject, ORKTaskViewControllerDeleg
             return instantiateActivityIntroductionStepViewController(for: schedule, step: step, taskRef: taskRef)
         }
         
+        // If the default view controller for this step is an `ORKInstructionStep` (and not a subclass)
```
Erin-Mounts: ORKInstructionStep --> ORKInstructionStepViewController

Erin-Mounts (approved):
a few nits, otherwise looks good. merge when ready

BridgeAppSDK/SBARoundedButton.swift
syoung-smallwisdom: ResearchKit

BridgeAppSDK/SBARoundedButton.swift
syoung-smallwisdom: It's there as a "soft marker" for indicating that the view is "hidden". ResearchKit has different places where it hides the continue button by setting alpha to `0` rather than using `isHidden`. This is a work-around to allow for a consistent continue button.

syoung-smallwisdom:
Note: I forgot to remove the references to the memory task (which is now deprecated) in the unit tests. Working on fixing that now.

PR #248, opened 2017-06-07
Bug/removes imageview twice

dephillipsmichael:
Fixed minor bug where imageview will get removed twice on return to screen if there is no instruction image

### DIFF
--- a/BridgeAppSDK/SBAInstructionStepViewController.swift
+++ b/BridgeAppSDK/SBAInstructionStepViewController.swift
@@ -109,7 +109,7 @@ open class SBABaseInstructionStepViewController: ORKStepViewController {
         if let image = self.image {
             self.imageView.image = image
         }
-        else {
+        else if self.imageView != nil {
             self.imageView.removeFromSuperview()
         }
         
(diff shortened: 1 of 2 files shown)


dephillipsmichael:
@syoung-smallwisdom not sure why i couldn't add you as a reviewer, but this PR is for you

BridgeAppSDK/SBAInstructionStepViewController.swift
```diff
@@ -109,7 +109,7 @@ open class SBABaseInstructionStepViewController: ORKStepViewController {
         if let image = self.image {
             self.imageView.image = image
         }
-        else {
+        else if self.imageView != nil {
```
syoung-smallwisdom: If you want to change this to possibly be nil, also change the declaration to optional. Currently, this is defined as *expected* to exist.

````
    @IBOutlet public weak var imageView: UIImageView?
````

syoung-smallwisdom (changes requested):

dephillipsmichael:
FTFY @syoung-smallwisdom

syoung-smallwisdom (approved):
looks good, except you will have merge failures b/c of some changes I made to the instruction step. Sorry about that. Please see if they resolve your issues.

dephillipsmichael:
This PR is actually not relevant anymore since you made the imageView optional.  :)

PR #254, opened 2017-06-09
Adding Config Options to BridgeInfo

joshbruhin:
Adding options to show and hide certain UI elements in new abstracted UI project


Erin-Mounts:
tests failing to run--from Travis:
[21:53:32]: ▸ ❌  /Users/travis/build/Sage-Bionetworks/BridgeAppSDK/BridgeAppSDKTests/SBAScheduledActivityManagerTests.swift:951:7: type 'TestScheduledActivityManager' does not conform to protocol 'SBABridgeInfo'

syoung-smallwisdom (changes requested):
I'd prefer to keep this kind of appearance management out of bridgeInfo, which is tied pretty tightly to how *our* server interacts with the model.  

I'd prefer to see something like we are doing with `UIColor` where there is an extension with class vars in BridgeAppSDK and overrides in the app.

An extension on `ORKStepViewController` perhaps?

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -91,6 +91,32 @@ public protocol SBABridgeInfo: SBASharedAppInfo, SBBBridgeInfoProtocol {
      do not check for a text user.
     */
     var disableTestUserCheck: Bool { get }
+    
+    /**
+     Show a back button in the navigation bar of ORKTaskViewController.
+     */
+    var showBackButtonInNavBar: Bool { get }
```
syoung-smallwisdom, line 98: since this is a question, should be `shouldShowBackBarInNavBar`.  Additionally, since it's one or the other (nav bar or footer) then an enum would be more appropriate.  Also, why are these defined at this level at all since there is no code at this level that references them?

joshbruhin:
Ok, will do

syoung-smallwisdom (changes requested):

BridgeAppSDK/SBABridgeInfo.swift
joshbruhin: I thought it could be that both are shown. I've seen that in other RK projects we've worked on and I can see how a designer might think that's a good idea. 

I thought it would be handy to put these options in a .plist, and I thought (mistakenly) BridgeInfo.plist was the right place.

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -91,6 +91,32 @@ public protocol SBABridgeInfo: SBASharedAppInfo, SBBBridgeInfoProtocol {
...
+    var showBackButtonInNavView: Bool { get }
+
+    /**
+     Place the navigationView in ORKStepViewController at bottom of screen, versus
+     embedded as the tableView's footerView.
+     */
+    var useStickyNavigationView: Bool { get }
+    
+    /**
+     Show a drop shadow above the navigationView in an ORKStepViewController
```
Erin-Mounts, line 112: missing '.' at end

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -91,6 +91,32 @@ public protocol SBABridgeInfo: SBASharedAppInfo, SBBBridgeInfoProtocol {
...
+     */
+    var showNavigationViewShadow: Bool { get }
+    
+    /**
+     Show a progresView at the top of the headerView in an ORKStepViewController
```
Erin-Mounts, line 117: '.'

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -144,6 +170,13 @@ extension SBAInfoManager: SBABridgeInfo {
         return (0,0)
     }
     
+    fileprivate func interfaceConfig() -> Dictionary<AnyHashable, Any>? {
+        if let configDict = self.plist["interfaceConfig"] as? Dictionary<AnyHashable, Any> {
```
Erin-Mounts, line 174: use #keyPath(interfaceConfig) instead of string literal

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -188,6 +221,40 @@ extension SBAInfoManager: SBABridgeInfo {
         return self.plist["disableTestUserCheck"] as? Bool ?? false
     }
 
+    public var showBackButtonInNavBar: Bool {
+        if let interfaceConfig = interfaceConfig() {
+            return interfaceConfig["showBackButtonInNavBar"] as? Bool ?? false
```
Erin-Mounts, line 226: #keyPath(...)

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -188,6 +221,40 @@ extension SBAInfoManager: SBABridgeInfo {
...
+        }
+        return false
+    }
+
+    public var showBackButtonInNavView: Bool {
+        if let interfaceConfig = interfaceConfig() {
+            return interfaceConfig["showBackButtonInNavView"] as? Bool ?? false
```
Erin-Mounts, line 233: #keyPath(...)

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -188,6 +221,40 @@ extension SBAInfoManager: SBABridgeInfo {
...
+        }
+        return false
+    }
+
+    public var useStickyNavigationView: Bool {
+        if let interfaceConfig = interfaceConfig() {
+            return interfaceConfig["useStickyNavigationView"] as? Bool ?? false
```
Erin-Mounts, line 240: #keyPath(...)

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -188,6 +221,40 @@ extension SBAInfoManager: SBABridgeInfo {
...
+        }
+        return false
+    }
+
+    public var showNavigationViewShadow: Bool {
+        if let interfaceConfig = interfaceConfig() {
+            return interfaceConfig["showNavigationViewShadow"] as? Bool ?? false
```
Erin-Mounts, line 247: #keyPath(...)

BridgeAppSDK/SBABridgeInfo.swift
```diff
@@ -188,6 +221,40 @@ extension SBAInfoManager: SBABridgeInfo {
...
+        }
+        return false
+    }
+    
+    public var showProgressView: Bool {
+        if let interfaceConfig = interfaceConfig() {
+            return interfaceConfig["showProgressView"] as? Bool ?? false
```
Erin-Mounts, line 254: #keyPath(...)

Erin-Mounts (changes requested):

BridgeAppSDK/SBABridgeInfo.swift
syoung-smallwisdom: The use of the plist for the BridgeInfo started as a work-around for AppCore and BridgeAppSDK not knowing about each other. When we split out some of the functionality into the factory framework and BridgeSDK, it allowed that flexibility.

That said, I'd like to move away from new development using this file.  It's useful for setting up the server config to allow us to extend what is supported by the server model, but it's not used for UI and I'd like to keep it agnostic to ResearchKit. :)

joshbruhin: Got it. Thanks for the background.

joshbruhin:
Cancelling PR. Alternate solution implement that does not involve BridgeAppSDK

PR #255, opened 2017-06-09
Add a navigable step as the intro step to allow for the “end early” survey

syoung-smallwisdom:
The daily check-in displays a survey asking the user why they selected "I can't do this now" which is uploaded to the server.  This needs to allow the intro step to navigate *AND* needs to cancel the update to the schedule to be "finished". This allows for that navigation.


Erin-Mounts (approved):
looks good

PR #258, opened 2017-06-13
More profile navigation

Erin-Mounts:
Support for html navigation by having the profile table item know how to find & return its html and its url.


syoung-smallwisdom (approved):
looks good

PR #259, opened 2017-06-14
open_method

dephillipsmichael:
Switched to open method to adjust the fields in registration steps


syoung-smallwisdom (approved):
Oops. Looks good

PR #261, opened 2017-06-19
Study participant profile item

Erin-Mounts:


Erin-Mounts:
requires https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/156

BridgeAppSDK/Resources/ProfileItems.json
```diff
@@ -0,0 +1,46 @@
...
+                 "itemType": "String"
+                 },
+                 {
+                 "profileKey": "participantLastName",
+                 "classType": "StudyParticipantProfileItem",
+                 "sourceKey": "lastName",
+                 "itemType": "String"
+                 },
+                 {
+                 "profileKey": "participantEmail",
```
syoung-smallwisdom, line 16: Can we call this "participantUsername" instead? The overloading on the term "email" is really confusing. :)

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -523,7 +557,143 @@ open class SBAUserDefaultsProfileItem: SBAProfileItemBase {
...
+        
+        studyParticipant.setValue(setValue, forKeyPath: sourceKey)
+        
+        // save the change to Bridge
+        SBABridgeManager.updateParticipantRecord(studyParticipant) { (_, _) in }
+    }
+}
+
+
+open class SBAStudyParticipantCustomAttributesProfileItem: SBAProfileItemBase {
```
syoung-smallwisdom, line 655: Can you add an example for this to the json? phone and contact email would both be of this type, yeah?

BridgeAppSDK/SBAProfileManager.swift
```diff
@@ -113,13 +113,20 @@ open class SBAProfileManager: SBADataObject, SBAProfileManagerProtocol {
      */
     public static let shared: SBAProfileManagerProtocol? = {
         SBABridgeManager.addResourceBundleIfNeeded()
-        guard let json = SBAResourceFinder.shared.json(forResource: SBAProfileItemsJSONFilename),
-                let sharedProfileManager = SBAClassTypeMap.shared.object(with:json, classType:SBAProfileManagerClassType) as? SBAProfileManagerProtocol
-            else {
-                // The SBAUser will default to the profile manager if available, but if not will fall-back
-                // to older methods for storing information.
-                return SBAProfileManager(dictionaryRepresentation: [:])
+        let bundles = SBAInfoManager.shared.resourceBundles.reversed()
```
syoung-smallwisdom, line 116: I am surprised that this doesn't put you into an infinite loop of wacky madness. The values in `resourceBundles` are changed during set up of the bridge connection to include BridgeAppSDK. Something to watch out for if it doesn't seem to be behaving as expected.

syoung-smallwisdom (approved):
Couple of questions. Otherwise, looks good. Merge when ready.

BridgeAppSDK/Resources/ProfileItems.json
Erin-Mounts: I worry participantUsername would be as confusing since we used to call that signUp field username, but was deprecated in favor of email since it must be an email address...

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: custom attributes are defined at the app level, so I could put it in the sample app I suppose but we haven't really been maintaining that so it needs a lot of TLC I think, which I was hoping to postpone until we get these dang pharma apps out. phone is of this type, as are birthdate and ethnicity in Journey Pro, but the email address in the StudyParticipant object is the contact email--it starts out with the email you signed up with, but you can change it without affecting the account identifier email, which never changes (and is stored completely separately by the app).

BridgeAppSDK/SBAProfileManager.swift
Erin-Mounts: seems to be working fine.

BridgeAppSDK/Resources/ProfileItems.json
syoung-smallwisdom: ok

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -121,6 +125,10 @@ extension SBAProfileItem {
         case SBAProfileTypeIdentifier.dictionary, SBAProfileTypeIdentifier.array:
             return (val as? SBAJSONObject)?.jsonObject() as? SBBJSONValue
             
+        case SBAProfileTypeIdentifier.set:
+            guard let set = val as? Set<AnyHashable> else { return nil }
+            return (Array(set) as SBAJSONObject).jsonObject() as? SBBJSONValue
```
syoung-smallwisdom, line 130: Note to self: Would be nice to merge the `SBAJSONObject` and `SBBJSONValue` protocols. ;)

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -523,7 +557,170 @@ open class SBAUserDefaultsProfileItem: SBAProfileItemBase {
...
-open class SBAFullNameProfileItem: SBAKeychainProfileItem, SBANameDataSource {
+enum SBAProfileParticipantSourceKey: String {
+    case firstName
+    case lastName
+    case email
+    case externalId
+    case notifyByEmail
+    case sharingScope
+    case dataGroups
+}
```
syoung-smallwisdom, line 568: Are these defined here rather than as obj-c enum strings b/c you explicitly want them *not* to extend?

Erin-Mounts: Exactly, yes.

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -523,7 +557,170 @@ open class SBAUserDefaultsProfileItem: SBAProfileItemBase {
...
+    
+    override open func storedValue(forKey key: String) -> Any? {
+        guard let studyParticipant = SBAStudyParticipantProfileItem.studyParticipant
+            else {
+                assertionFailure("Attempting to read \(key) (\(profileKey)) on nil SBBStudyParticipant")
+                return nil
+        }
+        guard !(key.hasPrefix("attributes.")) else {
+            let pathComponents = key.components(separatedBy: ".")
+            let attributeComponents = pathComponents[1...pathComponents.count - 1]
```
syoung-smallwisdom: `let attributeComponents = pathComponents[1..<pathComponents.count]` is equivalent and reads a bit easier.

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -523,7 +557,170 @@ open class SBAUserDefaultsProfileItem: SBAProfileItemBase {
...
+            guard attributeComponents.count > 0 else {
+                assertionFailure("Error reading \(key) (\(profileKey)): \(key) is not a valid SBBStudyParticipant keypath")
+                return nil
+            }
+            let attributeKey = attributeComponents.joined(separator:".")
```
syoung-smallwisdom: This bit is super confusing, and don't really understand what it is used for since the profile item can be explicitly defined as a custom attributes profile item. (By the way, why does this inherit from that one?)

Also, if this is necessary, can be simplified with:

````
// special-case handling for an attribute to call through to the superclass implementation
if let attributeKey = key.parseSuffix(prefix: "attributes", separator:".") {
    return super.storedValue(forKey: attributeKey)
}
````

File: BridgeAppSDK/SBAProfileItem.swift (hunk 6 above)
Comment (Erin-Mounts) [reply BridgeAppSDK/SBAProfileItem.swift]: thanks

File: BridgeAppSDK/SBAProfileItem.swift [hunk 8]
```diff
@@ -523,7 +557,170 @@ open class SBAUserDefaultsProfileItem: SBAProfileItemBase {
...
+        return storedVal
+    }
+    
+    override open func setStoredValue(_ newValue: Any?) {
+        guard let studyParticipant = SBAStudyParticipantProfileItem.studyParticipant
+            else {
+                assertionFailure("Attempting to set \(sourceKey) (\(profileKey)) on nil SBBStudyParticipant")
+                return
+        }
+        guard !(sourceKey.hasPrefix("attributes.")) else {
```
syoung-smallwisdom: See comment above. This is confusing. Sorry.

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -523,7 +557,170 @@ open class SBAUserDefaultsProfileItem: SBAProfileItemBase {
...
+            }
+        default:
+            // the rest are String, and anything can be converted to String
+            break
+        }
+        
+        studyParticipant.setValue(setValue, forKeyPath: sourceKey)
+        
+        // save the change to Bridge
+        SBABridgeManager.updateParticipantRecord(studyParticipant) { (_, _) in }
```
syoung-smallwisdom, line 650: Does this fire a server update? Seems a bit chatty.

syoung-smallwisdom (changes requested):
The bit with the attributes key is really confusing. Could you look at rewrite? Thanks.

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: I put this in (with this inheritance) so that a custom attribute key (preferredName) could fall back to a standard StudyParticipant key (firstName). I'll apply the simplification.

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: I kind of knew it would be... sorry

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: It will if the app is authenticated (has logged in and has a session token, not necessarily still valid). It won't during onboarding. I'm hoping edits to profile items stored here will be infrequent after that, and usually not done a bunch in a row. If that turns out not to be the case we can look at some kind of delay to gather a series of edits before firing off an update to the server, but this is simpler, easier to maintain, and more foolproof as far as making sure the updates all get to the server.

syoung-smallwisdom (approved):
looks good

PR #262, opened 2017-06-20
Survey animation

syoung-smallwisdom:
Note: this makes an older version of BridgeAppSDK that repurposes the detail on the last page as a copyright unavailable. Will require updating both elevateMS and JourneyPro once pushed.


Erin-Mounts (approved):
looks good

PR #265, opened 2017-06-24
Don’t allow requesting activities before join date

Erin-Mounts:
…except for testing


BridgeAppSDK/SBABridgeManager.m
```diff
@@ -246,8 +246,25 @@ + (void)fetchChangesToScheduledActivities:(NSArray <SBBScheduledActivity *> *)sc
...
+    _allowScheduleRequestsBeforeStartDate = allow;
+}
+
 + (void)fetchScheduledActivitiesFrom:(NSDate *)scheduledFrom to:(NSDate *)scheduledTo
                           completion:(SBABridgeManagerCompletionBlock)completionBlock {
+    if (!self.allowScheduleRequestsBeforeStartDate) {
+        // pin the scheduledFrom date to be no earlier than when the participant joined the study
+        NSDate *participantStartDate = SBAAppDelegate.shared.currentUser.createdOn;
+        NSComparisonResult order = [scheduledFrom compare:participantStartDate];
+        scheduledFrom = (order == NSOrderedAscending) ? participantStartDate : scheduledFrom;
```
syoung-smallwisdom: What about a class override of the user instead? Then use MockUser?

syoung-smallwisdom (approved):
Looks fine. A bit circuitous if you can find a good way to avoid directly accessing currentUser.  This wrapper was really intended to simplify the accees points by consolidating them. It wasn't really intended to need testing.

BridgeAppSDK/SBABridgeManager.m
Erin-Mounts: Not sure I'm understanding your suggestion, but I'm not seeing a straightforward way to follow it, mainly because SBABridgeManager consists entirely of class methods so there's no instance to set overrides on. In hindsight, I could have done a more straightforward setup with a testing global for the createdOn date and a class method to look at that and fall back to getting it from the current user if it's not set.

Looking at the existing tests for this again, though, I can see your point that they're basically pointless now that it's calling a single BridgeSDK method instead of making multiple calls to fake it. Would it make more sense just to remove them and not worry about overrides for testing here, or were they testing more than that? 

I see I also ought to add some code here to make sure the scheduledFrom and scheduledTo dates are in ascending order, and could get the start date directly from SBAUser.shared instead of going through the app delegate if you think that would be preferable (or even pick it directly out of the signUp response in SBABridgeManager and store it in a local static variable).

syoung-smallwisdom: Yes, those test were there to check that my work-around was merging the schedule objects since it was possible to get today's schedules using both calls. They can be deleted.

PR #266, opened 2017-06-25
Feature/performance

dephillipsmichael:
Added performance screen to BridgeAppSDK in a way that is easy to override the UI and functionality of the ViewController


BridgeAppSDK/SBAAppDelegate.swift
```diff
@@ -83,6 +83,11 @@ public let SBAMainStoryboardName = "Main"
         return UIApplication.shared.delegate as? SBAAppDelegate
     }
     
+    open var studyDuration: TimeInterval {
+        let oneYear = TimeInterval(60 * 60 * 24 * 365)
```
syoung-smallwisdom, line 87: Use a calendar. This works for 2017 but not 2018. ;)  Also, I'd suggest changing this to DateComponents instead since the rule might be "3 months" etc.

BridgeAppSDK/SBAPerformanceDataSource.swift
```diff
@@ -0,0 +1,169 @@
...
+    case myTrends
+    case count
+}
+
+public enum SBAPerformanceStudyProgressCellType: Int {
+    case timeLeftInStudy
+    case completedActivities
+}
+
+open class SBAPerformanceDataSource {
```
syoung-smallwisdom, line 24: I don't really see the value that this class is adding. It's not really a data source b/c it's not a protocol with a concrete implementation.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+    }
+    
+    // SBAScheduledActivityManagerDelegate method
+    open func reloadFinished(_ sender: Any?) {
+        self.tableView.reloadData()
+    }
+    
+    open func registerCells() {
+        // Register UITableViewCells
+        for cell in blankPerformanceCells {
```
syoung-smallwisdom, line 85: blankPerformanceCells?

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+            
+        case .myTrends:
+            let cell = tableView.dequeueReusableCell(withIdentifier: SBAPerformanceTrendsGraphTableViewCell.staticCellName, for: indexPath) as! SBAPerformanceTrendsGraphTableViewCell
+            self.configureMyTrendsCell(cell: cell, for: indexPath)
+            return cell
+        
+        default: break
+        }
+        
+//        cell.layoutIfNeeded() // <- added to fix bug where user had to scroll before correct autolayout height was calculated
```
syoung-smallwisdom, line 164: Then why is it commented out? ;)

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+
+import Foundation
+
+open class SBAPerformanceViewController: UITableViewController, SBAScheduledActivityManagerDelegate {
+    
+    open var activityManager : SBABaseScheduledActivityManager {
+        return _activityManager
+    }
+    lazy fileprivate var _activityManager : SBABaseScheduledActivityManager = {
+        return SBAScheduledActivityManager(delegate: self)
```
syoung-smallwisdom, line 42: Why not use the base class for the default implementation since you don't gain anything from this.  Also, if you refactor `SBABaseScheduledActivityManager` to use a a start/end date rather than daysBehind / daysAhead (deprecate those?) then you can set it up to show you the whole past or a range, as needed.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import Foundation
+
+open class SBAPerformanceViewController: UITableViewController, SBAScheduledActivityManagerDelegate {
```
syoung-smallwisdom, line 36: I think a better implementation of the architecture for this would be to use a data source that is a protocol with a general-use implementation (extension on  `SBABaseScheduledActivityManager`?). Otherwise, for expediency in getting this to production, you could just get ride of the data source and implement as a part of this controller.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+    }
+    
+    open func registerCells() {
+        // Register UITableViewCells
+        for cell in blankPerformanceCells {
+            self.tableView.register(UINib(nibName: cell.cellName, bundle: Bundle(for: object_getClass(cell))), forCellReuseIdentifier: cell.cellName)
+        }
+        
+        // Register UITableViewHeaderFooterViews
+        self.tableView.register(UINib.init(nibName: SBAPerformanceSectionHeader.staticCellName, bundle: Bundle(for: SBAPerformanceSectionHeader.self)), forHeaderFooterViewReuseIdentifier: SBAPerformanceSectionHeader.staticCellName)
```
syoung-smallwisdom, line 90: It's one thing to include hardcoded "it only uses this set of magic values" stuff in an app. For a class that is intended to be general-purpose, this kind of thing can make overriding messy. i.e., What happens when someone wants to use a header that inherits from `SBAPerformanceSectionHeader` rather than using the header design included here?

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+            switch cellType {
+                
+            case .timeLeftInStudy:
+                let cell = tableView.dequeueReusableCell(withIdentifier: SBAPerformanceProgressLinePlotTableViewCell.staticCellName, for: indexPath) as! SBAPerformanceProgressLinePlotTableViewCell
+                cell.titleLabel.text = Localization.localizedString("SBA_STUDY_PROGRESS_TIME_LEFT_TITLE")
+                cell.detailsLabel.attributedText = attributedTimeLeftStr()
+                self.configureStudyProgressTimeLeftCell(cell: cell, for: indexPath)
+                return cell
+                
+            default: //.completedActivities
```
syoung-smallwisdom, line 129: Use `.completedActivities` explicitly.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+    }
+    
+    override open func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        
+        guard let sectionType = performanceDataSource.sectionType(for: indexPath.section) else {
+            return self.cellForCustomSection(at: indexPath)
+        }
+        
+        switch sectionType {
+        case .studyProgress:
```
syoung-smallwisdom, line 118: Break this into a separate method. First, large blocks of nested code is difficult to read (and easy to mess up later). Second, that will allow override.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+        let average = total / scoreArray.count
+        
+        if (average < min) {
+            min = 0
+        }
+        if ((average > 100) && (average < 200)) {
+            max = 200
+        } else if (average > 200) {
+            max = average
+        }
```
syoung-smallwisdom, line 232: Um, no. Create a struct with min/max/average and return it in a method that calculates those values for a given activity.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+        
+        cell.leftLegendLabel.text = "\(min)"
+        cell.rightLegendLabel.text = "\(max)"
```
syoung-smallwisdom, line 235: I take that back. Also, the struct should return a formatter that is appropriate to the data.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+    }
+    
+    /**
+     * Styles the time left in study progress cell with the title,
+     * time left in months and days string in the details,
+     * and a progress line plot that is filled up based on time through the study
+     */
+    open func configureStudyProgressTimeLeftCell(cell: SBAPerformanceProgressLinePlotTableViewCell, for indexPath: IndexPath)
+    {
+        let appDelegate = UIApplication.shared.delegate as! SBAAppDelegate
```
syoung-smallwisdom, line 286: Danger, Will Robinson! No. Don't do this, please.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+        
+        // Convert study start and end dates to a normalized value between 0.0 and 1.0
+        let calendar = Calendar.current
+        let now = Date().timeIntervalSince1970
+        let normalizedValue = Float(now - startDate.timeIntervalSince1970) / Float(endDate.timeIntervalSince1970 - startDate.timeIntervalSince1970)
+        
+        // Calculate the number of months between start and end date and set that
+        // as the line plot's progress max
+        let monthsYearsBetweenStudyDates =  calendar.dateComponents([.year, .month], from: startDate, to: endDate)
+        let lengthOfStudyInMonths = (1 + /* +1 for including current month */(monthsYearsBetweenStudyDates.month ?? 0)) + ((monthsYearsBetweenStudyDates.year ?? 0) * 12) // 12 months in a year
```
syoung-smallwisdom, line 299: Ok, first, if you are using the current calendar and not the Gregorian calendar then you shouldn't assume the number of months is 12. Second, just get the months. oh yeah, and if you used `DateComponents` as suggested above, this can be calculated from that.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+        // Apply the value of 0 to 1 to the study's duration
+        cell.linePlotView.progress = Int(Float(lengthOfStudyInMonths) * normalizedValue)
+        
+        // Set the legend text, the study start month, middle month, and study end month
+        let monthStrs = DateFormatter().shortMonthSymbols!
+        cell.leftLegendLabel.text = monthStrs[(calendar.dateComponents([.month], from: startDate).month! - 1)] // -1 for 0 array indexing
+        let midDate = calendar.date(byAdding: .month, value: lengthOfStudyInMonths / 2, to: startDate)!
+        cell.midLegendLabel.text = monthStrs[(calendar.dateComponents([.month], from: midDate).month! - 1)] // -1 for 0 array indexing
+        let month = calendar.dateComponents([.month], from: endDate).month!
+        cell.rightLegendLabel.text = monthStrs[month - 1] // - 1 for 0 array indexing
```
syoung-smallwisdom, line 311:
````
let monthFormatter = DateFormatter()
monthFormatter.setLocalizedDateFormatFromTemplate("MMM")
cell.leftLegendLabel.text = monthFormatter.string(from: startDate)
let midDate = calendar.date(byAdding: .month, value: lengthOfStudyInMonths / 2, to: startDate)!
cell.midLegendLabel.text = monthFormatter.string(from: midDate)
cell.rightLegendLabel.text = monthFormatter.string(from: endDate)
````

File: BridgeAppSDK/SBAPerformanceViewController.swift [hunk 15]
```diff
@@ -0,0 +1,562 @@
...
+        let normalizedValue = Float(now - startDate.timeIntervalSince1970) / Float(endDate.timeIntervalSince1970 - startDate.timeIntervalSince1970)
+        
+        // Calculate the number of months between start and end date and set that
+        // as the line plot's progress max
+        let monthsYearsBetweenStudyDates =  calendar.dateComponents([.year, .month], from: startDate, to: endDate)
+        let lengthOfStudyInMonths = (1 + /* +1 for including current month */(monthsYearsBetweenStudyDates.month ?? 0)) + ((monthsYearsBetweenStudyDates.year ?? 0) * 12) // 12 months in a year
+        cell.linePlotView.progressMax = lengthOfStudyInMonths
+        
+        // Apply the value of 0 to 1 to the study's duration
+        cell.linePlotView.progress = Int(Float(lengthOfStudyInMonths) * normalizedValue)
```
syoung-smallwisdom, line 303: Your comment and the "progress" don't match. There are no `Int` values between 0 and 1.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+    }
+    
+    /**
+     Styles the time left in study string with large blue numbers and
+     centered month and day descriptors
+     */
+    open func attributedTimeLeftStr() -> NSMutableAttributedString {
+        let formatter = DateComponentsFormatter()
+        formatter.allowedUnits = [.month, .day]
+        formatter.unitsStyle = .full
```
syoung-smallwisdom, line 339: Where is this used?

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+        var dateComponents = DateComponents()
+        let (months, days) = timeUntilStudyEnds()
+        dateComponents.month = months
+        dateComponents.day = days
+        let timeUntilStudyEndsStr = String(format: "%d MONTHS %d DAYS", months, days)
```
syoung-smallwisdom, line 344: localize

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+     centered month and day descriptors
+     */
+    open func attributedTimeLeftStr() -> NSMutableAttributedString {
+        let formatter = DateComponentsFormatter()
+        formatter.allowedUnits = [.month, .day]
+        formatter.unitsStyle = .full
+        var dateComponents = DateComponents()
+        let (months, days) = timeUntilStudyEnds()
+        dateComponents.month = months
+        dateComponents.day = days
```
syoung-smallwisdom, line 343: Not seeing where the date components is being used, either.

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+        let timeUntilStudyEndsStr = String(format: "%d MONTHS %d DAYS", months, days)
+        let attributedString = NSMutableAttributedString(
+            string: timeUntilStudyEndsStr,
+            attributes: detailTextAttributes())
+        
+        let monthsRange = NSString(string: timeUntilStudyEndsStr).range(of: String(months))
+        attributedString.setAttributes(detailNumberAttributes(), range: monthsRange)
+        let daysRange = NSString(string: timeUntilStudyEndsStr).range(of: String(days))
+        attributedString.setAttributes(detailNumberAttributes(), range: daysRange)
+        return attributedString
```
syoung-smallwisdom, line 353: Localize.

````
let formatter = DateComponentsFormatter()
formatter.maximumUnitCount = 2
formatter.unitsStyle = .full
formatter.allowedUnits = [.year, .month, .day]
guard let timeUntilStudyEndsStr = formatter.string(from: Date(), to: endDate)?.localizedUppercase 
else { 
    return 
}

let attributedString = NSMutableAttributedString(
    string: timeUntilStudyEndsStr,
    attributes: detailTextAttributes())

var startIndex = 0
let searchString = NSString(string: timeUntilStudyEndsStr)
let length = searchString.length
while startIndex != NSNotFound, startIndex < length {
    let range = searchString.rangeOfCharacter(from: .decimalDigits, options: [], range: NSMakeRange(startIndex, length - startIndex))
    startIndex = range.location + range.length
    if startIndex != NSNotFound {
        attributedString.setAttributes(detailNumberAttributes(), range: range)
    }
}
````

File: BridgeAppSDK/SBAPerformanceViewController.swift [hunk 20]
```diff
@@ -0,0 +1,562 @@
...
+            let json = activity.clientData as? [Int]
+            let value = json?.last
+            if (value != nil) {
+                scores.append(value!)
+            } else {
+                // TODO: Remove this once scores are generated for every activity type.
+                // Right now this is a bit of a workaround to generate a score for activities that don't
+                // have them yet
+                // easieg 6/3/17
+                scores.append(0)
```
syoung-smallwisdom, line 420: Please change the comment

BridgeAppSDK/SBAPerformanceViewController.swift
```diff
@@ -0,0 +1,562 @@
...
+        var endDateWithCurrentMonth = calendar.date(byAdding: dateComponents as DateComponents, to: endDate)
+        var dayDif = calendar.dateComponents([.day], from: now, to: endDateWithCurrentMonth!)
+        if (dayDif.day! < 0) {
+            // This went too far back, so add 1 more month
+            // Admittedly a bit clunky, but handles edge cases
+            dateComponents.setValue((0-monthDif.month! + 1), forComponent: [.month])
+            endDateWithCurrentMonth = calendar.date(byAdding: dateComponents as DateComponents, to: endDate)
+            dayDif = calendar.dateComponents([.day], from: now, to: endDateWithCurrentMonth!)
+        }
+        return (monthDif.month!, dayDif.day!)
```
syoung-smallwisdom, line 447: tinker with Playground. This is seriously obfuscated.

syoung-smallwisdom (changes requested):
Stopped reviewing. Sorry. 

There is a lot of code here that could use more refactoring and clean up before trying to figure out what is genuinely useful in a shared code base. 

It might be more expedient just to copy/paste into JourneyPro for now and once we have some breathing room, we can look at setting some goals for generalized use.

dephillipsmichael:
I agree, at this point I should just add these classes into Journey Pro and continue to iterate on them to make them proper for BASDK.   But I'd like to keep this PR open for addressing the comments.

syoung-smallwisdom:
groovy

PR #267, opened 2017-06-27
bug/shadow_gradient

dephillipsmichael:
fix bug where gradient layer wasn't updating when sublayers are laid out


syoung-smallwisdom (approved):
looks good

PR #269, opened 2017-06-29
Fix signup caching bug

Erin-Mounts:


BridgeAppSDK/SBAUser.swift
```diff
@@ -41,7 +41,18 @@ import ResearchKit
...
-    static let shared = SBAUser()
+    static let shared: SBAUser = {
+        let user = SBAUser()
+        // If a previous install on this device was deleted, its keychain will still be around, but its
+        // userDefaults will have been wiped. In that case we want to make sure the old keychain is wiped
+        // so BridgeSDK doesn't see the old credentials and try to read/write from the old account during
+        // the onboarding/consent/signUp process.
+        if user.onboardingStepIdentifier == nil &&
+            !user.isLoginVerified {
+            user.resetKeychain()
```
syoung-smallwisdom: I think I'd rather put this somewhere it could be overridden. Maybe during appDelegate launch? There are a number of apps that don't use the onboarding step identifier. (Lilly, Smart4SURE, BiAffect)

syoung-smallwisdom (changes requested):

Erin-Mounts:
now requires https://github.com/Sage-Bionetworks/Bridge-iOS-SDK/pull/166

BridgeAppSDK/SBAAppDelegate.swift
```diff
@@ -87,6 +87,15 @@ public let SBAMainStoryboardName = "Main"
...
         // Override point for customization before application launch.
+        
+        // If a previous install on this device was deleted, its keychain will still be around, but its
+        // userDefaults will have been wiped. In that case we want to make sure the old keychain is wiped
+        // so BridgeSDK doesn't see the old credentials and try to read/write from the old account during
+        // the onboarding/consent/signUp process.
+        if currentUser.onboardingStepIdentifier == nil &&
+            !currentUser.isLoginVerified {
+            currentUser.resetStoredUserData()
+        }
```
syoung-smallwisdom: I actually meant to do something like this:

````
    /**
     Called on launch. This will reset the keychain if the user's stored information 
     indicates that they are logged out. This is necessary b/c deleting the app and reinstalling
     does not remove values stored in the keychain.
     */
    open func resetUserDataIfLoggedOut() {
        
        // If the user defaults keys all indicate logged out state
        // and *not* in the middle of onboarding, then reset the user.
        if !currentUser.isLoginVerified &&
            (currentUser.onboardingStepIdentifier == nil) &&
            !currentUser.isConsentVerified &&
            !currentUser.isRegistered {
            currentUser.resetStoredUserData()
        }
    }
````

The older apps to not use the `onboardingStepIdentifier` to track state, but they still need to track an onboarding state that is registered but *not* verified.

Erin-Mounts: ah, right. need to learn to let the coffee kick in before I try to code.

syoung-smallwisdom (approved):
Looks good

PR #273, opened 2017-07-05
Make property public for access by Elevate

joshbruhin:
Make 'tableView' on 'SBASignUpViewController' a public property so we can access it from Elevate


syoung-smallwisdom (approved):
whoops

PR #274, opened 2017-07-06
Client data profile items

Erin-Mounts:


BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -708,6 +709,366 @@ open class SBAStudyParticipantCustomAttributesProfileItem: SBAProfileItemBase {
...
+        return dict
+    }
+}
+
+public func < (lhs: SBAWhatAndWhen, rhs: SBAWhatAndWhen) -> Bool {
+    return (lhs.date as Date) < (rhs.date as Date)
+}
+
+public func == (lhs: SBAWhatAndWhen, rhs: SBAWhatAndWhen) -> Bool {
+    return (lhs.date as Date) == (rhs.date as Date)
```
syoung-smallwisdom: seems a bit obfuscated to only compare the dates. I'm not seeing where you are using equality that ignores the `value` property, but might want to  consider using `first(where:)` or `contains(where:)` instead, or if more appropriate, move the `replace(with newElements: [Element], where evaluate: (_ originalValue: Element, _ newValue: Element) ` utility extension out of the JourneyPro project into this one?

BridgeAppSDK/SBAProfileItem.swift
```diff
@@ -708,6 +709,366 @@ open class SBAStudyParticipantCustomAttributesProfileItem: SBAProfileItemBase {
     }
 }
 
+open class SBAWhatAndWhen: NSObject, Comparable {
+    static var valueKey: String { return #keyPath(value) }
+    static var dateKey: String { return #keyPath(date) }
+    static var isNewKey: String { return #keyPath(isNew) }
+    open var value: SBBJSONValue
+    open var date: NSDate
+    var isNew: Bool
```
syoung-smallwisdom, line 718: `isChanged`? Is this flag only set if the item is new or also if it needs to be updated on the server?

syoung-smallwisdom (approved):
Couple of nits. Otherwise, looks good.

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: The isNew flag indicates it's a new data point for this item in the local keychain cache, which gets cleared when it gets added to an appropriate SBBScheduledActivity.

BridgeAppSDK/SBAProfileItem.swift
Erin-Mounts: These are used when adding to a Set to prevent duplication, and when sorting an array by date. I can add value comparison as a secondary check. It's not useful for what I needed these for, but doesn't hurt either.

syoung-smallwisdom: Yes, please.

PR #275, opened 2017-07-06
make class func available outside the module

Erin-Mounts:
so app-specific view controllers can call it


syoung-smallwisdom (approved):
looks good

PR #276, opened 2017-07-08
Move custom Elevate UI to BridgeAppSDK

joshbruhin:
1) Move classes from 'Elevate' to ‘BridgeAppSDK’, rename as necessary. For example: 'ElevateStepViewController' -> 'SBAGenericStepViewController'
'ElevatePageStepViewController' -> 'SBAGenericPageStepViewController'
'ElevateStepTableData' -> 'SBAGenericStepDataSource'
'ElevateChoiceTableViewCell' -> 'SBAStepChoiceTableViewCell'
'ElevateStepHeaderView' -> 'SBAStepHeaderView'
'ElevateStepNavigationView' -> 'SBAStepNavigationView'
'ElevateStepProgressView' -> 'SBAStepProgressView'

2) Add ‘SBAVisualConsentStep’ for the consent scenes, which is a ‘ORKPageStep’. It creates a ‘SBAInstructionStep’ for each consent scene from the ‘ORKConsentDocument’. Also, modify ‘SBASurveyFactory’ to return the new ‘SBAVisualConsentStep’.

3) Extend ‘ORKFormStep’ and override ‘stepViewControllerClass()’ to return ‘SBAGenericStepViewController’

4) Extend ‘SBANavigationFormStep’ and override ‘stepViewControllerClass()’ to return ‘SBAGenericStepViewController’

5) Change the following View Controllers to inherit from our new ‘SBAGenericStepViewController’
‘SBAConsentSharingStepViewController’
‘SBAEmailVerificationStepViewController’
‘SBAExternalIDAssignStepViewController’
‘SBARegistrationStepViewController’

6) In ‘SBAScheduledActivityManager’, change ‘instantiateInstructionStepViewController’ to create a ‘SBAGenericStepViewController’ instead of a 'SBAInstructionStepViewController'.

7) Modify the ‘SBANavigationFormStep’ to return our new ‘SBAGenericStepViewController’ class.

8) Move ‘UIView+Layout’ from ‘Elevate’ to ‘BridgeAppSDK’ and document, rename and refactor.

9) Add ‘UIFont+BridgeKeyNames’ style class for use similar to the UIColor style class being used, both of which can be overridden by the app project.


syoung-smallwisdom:
Can you please look into ways to use SBASurveyFactory to override rather than moving UI elements into ResearchUXFactory-iOS ? I would like to restrict the UI elements that we use in this year's apps to BridgeAppSDK. Thanks!

joshbruhin:
@syoung-smallwisdom I moved everything out of 'ResearchUXFactory' and into 'BridgeAppSDK'. Should be good to go.

joshbruhin:
@syoung-smallwisdom getting Travis CI error: incorrect argument label in call (have 'identifier:', expected 'coder:')

Not sure what to make of this? the SBAVisualConsentStep has an initializer with 'identifier', and the required initializer with 'coder'. Any ideas?

BridgeAppSDK/SBAVisualConsentStep.swift
```diff
@@ -0,0 +1,88 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import UIKit
+
+open class SBAVisualConsentStep: ORKPageStep {
+    
+    private var consentDocument: ORKConsentDocument!
+    
+    public init(identifier: String, consentDocument: ORKConsentDocument) {
```
syoung-smallwisdom, line 41: ResearchKit requires that you implement init(identifier: String) or else the copy will fail. You need to define that as well. (This is why the tests are failing b/c ResearchKit does a *lot* of copying)

BridgeAppSDK/SBAGenericPageStepViewController.swift
```diff
@@ -0,0 +1,67 @@
...
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+
+import UIKit
+
+class SBAGenericPageStepViewController: ORKPageStepViewController, UIPageViewControllerDelegate {
```
syoung-smallwisdom: This is not generic. (Sorry) In fact, this is fairly specific to the one case where you are using it (to show consent flow). In the interest of time, just be explicit and call it `SBAVisualConsentStepViewController`. The general-purpose page step view controller defined in ResearchKit is a much more appropriate generic implementation because it lets the step determine what kind of view controller to vend (which may very well *not* be the generic one).

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+        }
+    }
+    
+    func updateAnswers(from result: ORKResult) {
+        
+        // iterate the provided results, find the corresponding ItemGroup for each, and save the answer
+        // provided in the results
+        
+        if let stepResult = result as? ORKStepResult {
+            if let results = stepResult.results as? Array<ORKQuestionResult> {
```
syoung-smallwisdom:
````
guard let stepResult = result as? ORKStepResult, 
    let results = stepResult.results as? [ORKQuestionResult] 
else {
    return nil
}
````

More appropriate in Swift to use early exit. Also, results in less nested curly braces which are hard to read.

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 4]
```diff
@@ -0,0 +1,614 @@
...
+                }
+            }
+        }
+    }
+    
+    
+    public func updateDefaults(_ defaults: NSMutableDictionary) {
+        
+        // TODO: Josh Bruhin, 6/12/17 - implement. this may require access to a HealthKit source.
+        // This has not yet been needed for the onboarding flow
```
syoung-smallwisdom: implement what?

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+                // Skipped forms report a "null" value for every item -- by skipping, the user has explicitly said they don't want
+                // to report any values from this form.
+                
+                var answer = ORKNullAnswerValue()
+                var answerDate = now
+                var systemCalendar = Calendar.current
+                var systemTimeZone = NSTimeZone.system
+                
+                // TODO: Josh Bruhin, 6/12/17 - implement skipped?
+                let skipped = false
```
syoung-smallwisdom: Cannot accept this for general purpose use without implementation that allows skipping steps. However, if the `itemGroup.answer` is nil, then isn't that effectively the same thing?

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+                
+                if let dateAnswerFormat = impliedAnswerFormat as? ORKDateAnswerFormat, let dateQuestionResult = result as? ORKDateQuestionResult {
+                    if let _ = dateQuestionResult.dateAnswer {
+                        let usedCalendar = dateAnswerFormat.calendar ?? systemCalendar
+                        dateQuestionResult.calendar = usedCalendar
+                        dateQuestionResult.timeZone = systemTimeZone
+                    }
+                } else if let numericAnswerFormat = impliedAnswerFormat as? ORKNumericAnswerFormat {
+                    if let numericQuestionFormat = result as? ORKNumericQuestionResult {
+                        numericQuestionFormat.unit = numericAnswerFormat.unit
```
syoung-smallwisdom: only set the unit if nil. also, don't nest if statements like this. hard to read.

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+                }
+                
+                let result = formItem.answerFormat?.result(withIdentifier: formItem.identifier, answer: answer)
+                let impliedAnswerFormat = formItem.answerFormat?.implied()
+                
+                if let dateAnswerFormat = impliedAnswerFormat as? ORKDateAnswerFormat, let dateQuestionResult = result as? ORKDateQuestionResult {
+                    if let _ = dateQuestionResult.dateAnswer {
+                        let usedCalendar = dateAnswerFormat.calendar ?? systemCalendar
+                        dateQuestionResult.calendar = usedCalendar
+                        dateQuestionResult.timeZone = systemTimeZone
```
syoung-smallwisdom: What about the date type? There are both date only and timestamp that get converted to an `ORKDateQuestionResult`?

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+                    
+                    if let itemGroup = itemGroup(with: formItem.identifier) {
+                        answer = itemGroup.answer
+                        answerDate = itemGroup.answerDate ?? now
+                        systemCalendar = itemGroup.calendar
+                        systemTimeZone = itemGroup.timezone
+                    }
+                }
+                
+                let result = formItem.answerFormat?.result(withIdentifier: formItem.identifier, answer: answer)
```
syoung-smallwisdom:
````
guard let result = ... else { continue }
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 9]
```diff
@@ -0,0 +1,614 @@
...
+                } else if let numericAnswerFormat = impliedAnswerFormat as? ORKNumericAnswerFormat {
+                    if let numericQuestionFormat = result as? ORKNumericQuestionResult {
+                        numericQuestionFormat.unit = numericAnswerFormat.unit
+                    }
+                }
+                
+                result?.startDate = answerDate
+                result?.endDate = answerDate
+                
+                qResults.append(result!)
```
syoung-smallwisdom: just above you aren't assuming there is a result but now you are forcing the optional?

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+            }
+            
+            parentResult.results = parentResult.results! + qResults
```
syoung-smallwisdom: No. First, using the `+` operator here compiles really slowly. Second, use `parentResult.addResult(result)` above instead. That method includes a check that you aren't adding a result with the same identifier more than once.

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+        if let formItems = formItems() {
+            var items: Array<ORKFormItem> = Array()
+            formItems.forEach({
+                if let _ = $0.answerFormat {
+                    items.append($0)
+                }
+            })
+            return items
+        }
+        return nil
```
syoung-smallwisdom:
````
return self.formItems()?.filter { $0.answerFormat != nil }
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 12]
```diff
@@ -0,0 +1,614 @@
...
+    
+    fileprivate func formItems() -> Array<ORKFormItem>? {
+        switch self.step {
+        case is ORKFormStep:
+            return (self.step as! ORKFormStep).formItems
+        case is ORKQuestionStep:
+            return (self.step as! ORKQuestionStep).formItems
+        default:
+            return nil
+        }
```
syoung-smallwisdom:
````
fileprivate func formItems() -> [ORKFormItem]? {
    guard let formStep = self.step as? SBAFormStepProtocol else { return nil }
    return formStep.formItems
}
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 13]
```diff
@@ -0,0 +1,614 @@
...
+    }
+    
+    fileprivate func populate() {
+        
+        if let items = formItems(), items.count > 0 {
```
syoung-smallwisdom: use a guard and exit early, please.

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+            
+            let singleSelectionTypes: Array<Int> = [ORKQuestionType.boolean.rawValue,
+                                                    ORKQuestionType.singleChoice.rawValue,
+                                                    ORKQuestionType.multipleChoice.rawValue,
+                                                    ORKQuestionType.location.rawValue]
```
syoung-smallwisdom:
````
let singleSelectionTypes: [ORKQuestionType] = [.boolean, .singleChoice, .multipleChoice, .location]
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 15]
```diff
@@ -0,0 +1,614 @@
...
+            
+            var newSection: SBAGenericStepTableSection? = nil
```
syoung-smallwisdom: use force optional here rather than everywhere you use it below. 

````
var newSection: SBAGenericStepTableSection!
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 16]
```diff
@@ -0,0 +1,614 @@
...
+                    else {
+                        
+                        newSection!.add(formItem: item)
+                        sections.append(newSection!)
+                    }
+                }
+                else {
+                    
+                    newSection!.title = item.text
+                    sections.append(newSection!)
```
syoung-smallwisdom: I don't understand why you are appending the same section to the sections array several times. This whole thing is really obfuscate and has a lot of duplicated code. That's the kind of thing that is hard to maintain and will often result in bugs introduced during maintenance.

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+            
+            for item in items {
+                
+                // In case no section available, create new one.
+                if newSection == nil {
+                    newSection = SBAGenericStepTableSection(sectionIndex: sections.count)
+                }
+                
+                if let answerFormat = item.answerFormat?.implied() {
+                    let multiCellChoices = singleSelectionTypes.contains(answerFormat.questionType.rawValue) && type(of: answerFormat) != ORKValuePickerAnswerFormat.self
```
syoung-smallwisdom:
````
let multiCellChoice = singleSelectionTypes.contains(answerFormat.questionType) && !(answerFormat is ORKValuePickerAnswerFormat)
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 18]
```diff
@@ -0,0 +1,614 @@
...
+    private var _title: String?
+    var title: String? {
+        get { return _title }
+        set (newValue) { _title = newValue?.uppercased(with: Locale.current) }
+    }
+    
+    var index: Int?
+    
+    public init(sectionIndex: Int) {
+        index = sectionIndex
```
syoung-smallwisdom: missing call to super.init()

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+    
+    open var itemGroups: Array<SBAGenericStepTableItemGroup> = Array()
+    
+    private var _title: String?
+    var title: String? {
+        get { return _title }
+        set (newValue) { _title = newValue?.uppercased(with: Locale.current) }
+    }
+    
+    var index: Int?
```
syoung-smallwisdom: This is never set (other than in the initializer). Use a `let` instead and don't make it optional.

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+    
+    /**
+     Add a new ORKFormItem, which results in the creation and addition of a new SBAGenericStepTableItemGroup to the section.
+     The ItemGroup essectially represents the FormItem and is reponsible for storing and providing answers for the FormItem
+     when a ORKStepResult is requested.
+     
+     @param   formItem    The ORKFormItem to add to the section
+     */
+    public func add(formItem: ORKFormItem) {
+        itemGroups.append(SBAGenericStepTableItemGroup(formItem: formItem, beginningRowIndex: itemCount()))
```
syoung-smallwisdom: Please add validation that the form items are unique.

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+
+open class SBAGenericStepTableItemGroup: NSObject {
+    
+    let formItem: ORKFormItem!
+    
+    var items: Array<SBAGenericStepTableItem> = Array()
+    var beginningRowIndex = 0
+    
+    // TODO: Josh Bruhin, 6/12/17 - implement multi-selection formItems. For now, set to singleSelection
+    var singleSelection: Bool = true
```
syoung-smallwisdom: Sorry, needs to support multiple selection. This is really only useful as a general-purpose UI if it supports the common survey types and multiple selection is really common.

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+    }
+    
+    /**
+     Returns the total count of all Items in this section.
+     @return    The total number of SBAGenericStepTableItems in this section
+     */
+    public func itemCount() -> Int {
+        var count = 0
+        itemGroups.forEach({ count += $0.items.count })
+        return count
```
syoung-smallwisdom:
````
return itemGroups.reduce(0, {$0 + $1.items.count})
````

`+=` compiles really slowly.

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 23]
```diff
@@ -0,0 +1,614 @@
...
+    
+    // TODO: Josh Bruhin, 6/12/17 - implement multi-selection formItems. For now, set to singleSelection
+    var singleSelection: Bool = true
+    
+    var answerDate: Date?
+    var calendar = Calendar.current
+    var timezone = TimeZone.current
+    
+    var defaultAnswer: AnyObject = ORKNullAnswerValue() as AnyObject
+    var _answer: AnyObject?
```
syoung-smallwisdom: `Any` and not `AnyObject`, isn't it? Arrays aren't `AnyObject` and you need arrays to support multiple selection.

Also, should this be private?

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+    
+    /**
+     Determine if the current answer is valid. Also checks the case where answer is required but one has not been provided.
+     @return    A Bool indicating if answer is valid
+     */
+    public var isAnswerValid: Bool {
+        
+        var valid = true
+        
+        // if answewr is NOT optional and it equals Null (ORKNullAnswerValue()), or is nil, then it's invalid
```
syoung-smallwisdom: answewr  -> answer

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+        if !formItem.isOptional, answer is NSNull || answer == nil {
+            valid = false
```
syoung-smallwisdom: just exit early.

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+        }
+        
+        // also check the value
+        if !formItem.answerFormat!.implied().isAnswerValid(answer) {
+            valid = false
+        }
+        return valid
```
syoung-smallwisdom:
````
return formItem.answerFormat?.implied().isAnswerValid(answer) ?? false
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 27]
```diff
@@ -0,0 +1,614 @@
...
+        var rowIndex = beginningRowIndex
+        
+        if let textChoiceAnswerFormat = formItem.answerFormat?.implied() as? ORKTextChoiceAnswerFormat {
+            
+            textChoiceAnswerFormat.textChoices.forEach({
+                let index = textChoiceAnswerFormat.textChoices.index(of: $0)!
+                let tableItem = SBAGenericStepTableItem(formItem: formItem, choiceIndex: index, rowIndex: rowIndex)
+                items.append(tableItem)
+                rowIndex += 1
+            })
```
syoung-smallwisdom:
````
self.items = textChoiceAnswerFormat.textChoices.enumerated().map { (index, _) -> SBAGenericStepTableItem in
    SBAGenericStepTableItem(formItem: formItem, choiceIndex: index, rowIndex: beginningRowIndex + index)
}
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 28]
```diff
@@ -0,0 +1,614 @@
...
+        } else {
+            let tableItem = SBAGenericStepTableItem(formItem: formItem, choiceIndex: 0, rowIndex: rowIndex)
+            items.append(tableItem)
```
syoung-smallwisdom: you aren't mutating the items once they are set.

`self.items = [tableItem]`

and define the items as forced optional.

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+            items[index].selected = selected
+        }
+        
+        // if we selected an item and this is a single-selection group, then we iterate
+        // our other items and de-select them
+        
+        if selected && singleSelection {
+            items.forEach({
+                if $0 != items[index] { $0.selected = false }
+            })
```
syoung-smallwisdom:
````
for (ii, item) in items.enumerated() {
    item.selected = (ii == index)
}
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 30]
```diff
@@ -0,0 +1,614 @@
...
+        }
+    }
+    
+    fileprivate func internalAnswer() -> AnyObject {
+        
+        if items.count > 0 {
+            
+            let answerFormat = items[0].formItem?.answerFormat
```
syoung-smallwisdom: Use guard and early exit

````
guard let answerFormat = items.first?.formItem?.answerFormat
else {
    return _answer ?? defaultAnswer
}
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift [hunk 31]
```diff
@@ -0,0 +1,614 @@
...
+            }
+        }
+        return _answer ?? defaultAnswer
+    }
+    
+    fileprivate func setInternalAnswer(_ answer: AnyObject) {
+        
+        if items.count > 0 {
+            
+            let answerFormat = items[0].formItem?.answerFormat
```
syoung-smallwisdom: guard and early exit

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
...
+                                item.selected = choiceNumber.intValue == selectedNumber.intValue
+                            }
+                        }
+                    }
+                }
+                
+            case is ORKMultipleValuePickerAnswerFormat:
+                
+                // TODO: Josh Bruhin, 6/12/17 - implement this answer format.
+                fatalError("setInternalAnswer for ORKMultipleValuePickerAnswerFormat not implemented")
```
syoung-smallwisdom: Sorry, this is used in some apps. You need to have a "not supported" case that can *not* use this step view controller for the cases where the UI doesn't support them and then be able to gracefully change to using the default UI from ResearchKit.

syoung-smallwisdom (changes requested):
There is a lot I wasn't able to get through. I can add more tomorrow.

BridgeAppSDK/SBAStepChoiceCell.swift
```diff
@@ -0,0 +1,327 @@
...
+    private let kTextFieldWidth: CGFloat = 150.0
+
+    override open func setupViews() {
+        
+        super.setupViews()
+        
+        textField.textAlignment = .center
+        textField.font = UIFont.textFieldFeaturedCellText
+        
+        // we don't want the label label
```
syoung-smallwisdom: labelLabel?

BridgeAppSDK/SBAGenericStepDataSource.swift
joshbruhin: Yeah, this section was one of the first things I worked on and I copied this code directly from RK ((void)buildSections in ORKFormStepVC), thinking it was safest to do exactly what RK was doing. I agree that it's not well written. I'll redo it.

BridgeAppSDK/SBAStepHeaderView.swift
```diff
@@ -0,0 +1,380 @@
...
+    private let kTopMargin: CGFloat = CGFloat(30.0).proportionalToScreenWidth()
+    private let kSideMargin: CGFloat = CGFloat(30.0).proportionalToScreenWidth()
+    private let kVerticalSpacing: CGFloat = CGFloat(20.0).proportionalToScreenWidth()
+    private let kBottomMargin: CGFloat = CGFloat(30.0).proportionalToScreenWidth()
+    private let kPromptBottomMargin: CGFloat = CGFloat(10.0).proportionalToScreenWidth()
+    private let kImageViewHeight: CGFloat = CGFloat(100.0).proportionalToScreenWidth()
+    private let kLearnMoreButtonHeight: CGFloat = 30.0
+    private let kLabelMaxLayoutWidth: CGFloat = {
+        return CGFloat(UIScreen.main.bounds.size.width - (2 * CGFloat(30.0).proportionalToScreenWidth()))
+    }()
```
syoung-smallwisdom, line 67: What happens to these when the screen rotates?

BridgeAppSDK/SBAStepHeaderView.swift
```diff
@@ -0,0 +1,380 @@
...
+            return shouldShowProgress && progressView.progress > 0
+        case is UIImageView:
+            return (view as! UIImageView).image != nil
+        case is UILabel:
+            return ((view as! UILabel).text?.characters.count ?? 0) > 0
+        case is UIButton:
+            return shouldShowLearnMore
+        default:
+            return true
+        }
```
syoung-smallwisdom: It is best to avoid forced casting. Unlike Obj-c, Swift is designed to use optional casting to determine whether or not something conforms to a type. It takes a bit of practice to learn to read it (especially coming from Java), but using the expected syntax makes for more consistent code which in turn is much easier to maintain. 

````
    guard view != nil else { return false }

    if let progressView = view as? SBAStepProgressView {
        return shouldShowProgress && progressView.progress > 0
    }
    else if let imageView = view as? UIImageView {
        return imageView.image != nil
    }
    else if let label = view as? UILabel {
        return (label.text?.characters.count ?? 0) > 0
    }
    else if let button = view as? UIButton, button == self.learnMoreButton {
        // If this is the learn more button then only show if used.
        return shouldShowLearnMore
    }

   return true
````

File: BridgeAppSDK/SBAStepProgressView.swift [hunk 36]
```diff
@@ -0,0 +1,156 @@
...
+ that is positioned under the progress bar to display current step and number of steps.
+ Views should set currentStep and totalSteps to cause progress to be displayed.
+ */
+
+open class SBAStepProgressView: UIView {
+    
+    /**
+     The current step in the current flow
+     */
+    public var currentStep: Int = 0 { didSet { progressChanged() } }
```
syoung-smallwisdom: Nit: Can you please put this on separate line? It's more consistent with use of `didSet` elsewhere.

BridgeAppSDK/SBAStepProgressView.swift
```diff
@@ -0,0 +1,156 @@
...
+     */
+    open func progressLineHeight() -> CGFloat {
+        return currentStep > 0 && totalSteps > 0 ? kProgressLineHeight : 0.0
+    }
+    
+    /**
+     The text of the label that is displayed directly under the progress bar
+     */
+    open func stringForLabel() -> String? {
+        // TODO: Josh Bruhin, 6/12/17 - localize string
```
syoung-smallwisdom: Please localize anything you are hoping to add to the common framework.

BridgeAppSDK/SBASurveyFactory.swift
```diff
@@ -226,3 +224,17 @@ open class SBASurveyFactory : SBABaseSurveyFactory {
...
+        return SBAGenericStepViewController.self
+    }
+}
+
+// Use our generic step view controller for any 'ORKFormStep'
+extension ORKFormStep {
+    override open func stepViewControllerClass() -> AnyClass {
+        return SBAGenericStepViewController.self
+    }
+}
```
syoung-smallwisdom: What about `ORKQuestionStep` and `SBANavigationQuestionStep`? Thought I saw code that could get the formItems array for an `ORKQuestionStep`? 

Note: these have different UI implementations in ResearchKit which is why, historically, we've had to use both.

Also, I'd prefer that you used the same pattern that I setup for an instruction step where the `SBAScheduledActivityManager` has an instantiate method that looks at the type of class that the step defaults to returning. It's hella kludgy but does allow for easier override for apps that are *not* designed to match elevateMS and JourneyPro and use their own stepViewController UI and keeps the overrides all in the same place using the same pattern.

BridgeAppSDK/UIView+Layout.swift
```diff
@@ -0,0 +1,342 @@
+//
+//  UIView+LayoutExtensions.swift
+//  ResearchUXFactory
+//
+//  Copyright © 2016 Sage Bionetworks. All rights reserved.
```
syoung-smallwisdom: copy/paste? copyright info is incorrect.

BridgeAppSDK/UIView+Layout.swift
```diff
@@ -0,0 +1,342 @@
...
+                
+                for constraint in constraints {
+                    
+                    // iOS automatically creates special types of constraints, like for intrinsicContentSize,
+                    // and we don't want these. So we make sure we have a 'NSLayoutConstraint' base class.
+                    if type(of: constraint) != NSLayoutConstraint.self {
+                        continue
+                    }
+                    
+                    if constraint.firstItem as! NSObject == self && constraint.firstAttribute == attribute && constraint.relation == relation {
```
syoung-smallwisdom: `SBAObjectEquality(constraint.firstItem, self)`

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,614 @@
+//
+//  SBAGenericStepDataSource.swift
+//  ResearchUXFactory
```
syoung-smallwisdom: BridgeAppSDK

BridgeAppSDK/SBAGenericStepDataSource.swift
syoung-smallwisdom: Thanks!  :)  And yes, unfortunately, I recognize it. (sigh)

BridgeAppSDK/SBAGenericStepDataSource.swift
joshbruhin: Got it. When it comes to the model, I generally followed patterns in RK; this is how they do it in most places. For populating and returning an ORKResult, I essentially copied the code. Again, thinking that was safest.

syoung-smallwisdom (changes requested):
Additional review comments.

Also, you appear to have a few changes in the factory that are conflicting with master.

And could you please add unit tests for the `SBAGenericStepDataSource`? Not having this tied directly to UI is great b/c it facilitates unit testing, but only if you actually include tests.

BridgeAppSDK/SBAGenericStepDataSource.swift
syoung-smallwisdom: Makes sense... but... (smile)... ResearchKit is pretty darn obtuse, and it's written in Obj-c. In Swift, things like this take a *really* long time to compile (seriously painful) and are best avoided. :)

BridgeAppSDK/SBAGenericStepDataSource.swift
joshbruhin: Copied directly form RK - (ORKStepResult *)result in ORKFormStepVC. I'll look into it...

syoung-smallwisdom: Thanks. It's probably fine, but... yeah. I still find bugs in RK so figured I'd ask. Although... this is a great reason to add unit tests. ;)

BridgeAppSDK/SBAGenericStepDataSource.swift
joshbruhin: Makes sense. There are a few sections in this class where I copied from RK, I'll have to make them swifty.

BridgeAppSDK/SBAGenericPageStepViewController.swift
joshbruhin: Agreed. Actually, this class is only necessary to show progress and hide the back bar button, which I haven't figured out how to do :( It doesn't need to return a VC if the step is defining it. So in that case, maybe it is generic and simply supports progress and maybe back button option?

syoung-smallwisdom: Yeah, it's a bit more complicated than that. You need to check that the view controller to be returned *should* be a generic *and* you need to check that the steps are a straight line and do not include navigation that would make your method of determining step count inaccurate. That's why the ResearchKit version of the task view controller progress asks the task rather than just counting steps.

For now, just explicitly use it for consent. :)

*OR* you can have `SBAGenericStepViewController` call the method on `self.step.task?.progress` (or something like that, can't remember what it's named and don't have time at the moment to code hunt.  That should work b/c internally the `ORKPageStep` is backed by a subclass of ORKOrderedTask.

BridgeAppSDK/SBAGenericStepDataSource.swift
joshbruhin: Sorry, reading default HealthKit values to update the model. In RK, this is all done in private methods -> (void)refreshDefaults in ORKFormStepVC. Anywhere in Bridge I can access this data? I've been looking, but no luck.

BridgeAppSDK/SBAGenericStepDataSource.swift
joshbruhin: Did you mean "only set the unit if *not* nil"?

BridgeAppSDK/SBAGenericStepDataSource.swift
syoung-smallwisdom: Ugh, no. This is intensionally left up to researchKit to handle. Sorry

BridgeAppSDK/SBAGenericStepDataSource.swift
syoung-smallwisdom: I mean, check that it's not already set:

````
else if let numericAnswerFormat = impliedAnswerFormat as? ORKNumericAnswerFormat,
    let numericQuestionFormat = result as? ORKNumericQuestionResult, 
    numericQuestionFormat.unit == nil {
        numericQuestionFormat.unit = numericAnswerFormat.unit
}
````

File: BridgeAppSDK/SBAGenericStepDataSource.swift (hunk 4 above)
Comment (joshbruhin) [reply BridgeAppSDK/SBAGenericStepDataSource.swift]: Can I submit a pull request to make the necessary functions public?

**Comment by syoung-smallwisdom** [comment]
Absolutely, we've forked. :)

**Comment by syoung-smallwisdom** [comment]
@joshbruhin Are you ready for me to review again? Github doesn't seem to have a way to mark that. Thanks.

**Comment by joshbruhin** [comment]
@syoung-smallwisdom I have two outstanding items: 1) a function to get profile values from HealthKit and update the model and 2) unit tests for the SBAGenericStepDataSource. Got distracted with a couple items in Elevate. Maybe we can review again and merge (hopefully) and I can do a new PR with these items? I realize this may not be possible if it affects other apps/projects...

**Comment by syoung-smallwisdom** [comment]
@joshbruhin Can you change health kit types to "not supported" so that they get displayed using the ResearchKit UI (for now) and then once they are supported, add here?

**Comment by joshbruhin** [comment]
@syoung-smallwisdom sounds do-able, I'll have a look...

**Comment by joshbruhin** [comment]
@syoung-smallwisdom SBAGenericStepVC now returns false when doesSupport() is called with a HK-related step. Took just a small change...filter based on answerFormat instead of answerFormat.implied()

File: BridgeAppSDK/SBAVisualConsentStep.swift (hunk 1 above)
Comment (syoung-smallwisdom) [reply BridgeAppSDK/SBAVisualConsentStep.swift:41]: @joshbruhin Ping

Comment (joshbruhin) [reply BridgeAppSDK/SBAVisualConsentStep.swift:41]: got it

Comment (joshbruhin) [reply BridgeAppSDK/SBAVisualConsentStep.swift:41]: Done

Comment (joshbruhin) [reply BridgeAppSDK/SBAVisualConsentStep.swift:41]: I'll check CI build status in a bit

File: BridgeAppSDK/SBAGenericStepViewController.swift [hunk 42]
```diff
@@ -0,0 +1,951 @@
...
+        // show or hide our navigation bar
+        navigationController?.setNavigationBarHidden(!shouldShowNavigationBar(), animated: false)
+        
+        setupViews()
+    }
+        
+    override open func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        
+        NotificationCenter.default.removeObserver(self)
```
syoung-smallwisdom: Sorry, missed this before. Could you please be specific about what notification is removed. Since this inherits from another class and b/c it is open, it can be seriously frustrating trying to track down why a notification isn't being received that you are expecting. Only to realize that one of the subclassing level was doing a general-purpose remove all. 

Thanks.

BridgeAppSDK/SBAGenericStepViewController.swift
```diff
@@ -0,0 +1,951 @@
...
+    }
+    
+    
+    // MARK: View lifecycle
+    
+    override open func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // register for keyboard notifications
+        NotificationCenter.default.addObserver(self, selector: #selector(self.keyboardNotification(notification:)), name: NSNotification.Name.UIKeyboardWillChangeFrame, object: nil)
```
syoung-smallwisdom: Add in `viewWillAppear`. The design of the learn more may include a full screen modal in which case this will not be removed in `viewWillDisappear` but not added again after the learn more modal goes away.

BridgeAppSDK/SBAGenericStepViewController.swift
```diff
@@ -0,0 +1,951 @@
...
+                }
+
+                if let placeholder = itemGroup?.formItem.placeholder {
+                    // populate the text field placeholder label
+                    textFieldCell.setPlaceholderText(placeholder)
+                }
+
+            }
+        } else {
+            if let choiceCell = cell as? SBAStepChoiceCell {
```
syoung-smallwisdom: `else if` rather than the double nested if statement.

BridgeAppSDK/SBAGenericStepViewController.swift
```diff
@@ -0,0 +1,951 @@
...
+    
+    open func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        return UITableViewAutomaticDimension
+    }
+    
+    // MARK: UITextField delegate
+    
+    public func textFieldDidBeginEditing(_ textField: UITextField) {
+        
+        if let customField = textField as? SBAStepTextField {
```
syoung-smallwisdom: use a guard

BridgeAppSDK/SBAGenericStepViewController.swift
```diff
@@ -0,0 +1,951 @@
...
+        tableView?.setContentOffset(CGPoint(x: 0.0, y: savedVerticalScrollOffet), animated: true)
+    }
+    
+    
+    public func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        
+        let textFieldText: NSString = (textField.text ?? "") as NSString
+        let textAfterUpdate = textFieldText.replacingCharacters(in: range, with: string)
+        
+        if let customTextField = textField as? SBAStepTextField {
```
syoung-smallwisdom: use a guard and early exit.

BridgeAppSDK/SBAGenericStepViewController.swift
```diff
@@ -0,0 +1,951 @@
...
+        let maxY = tableView!.contentSize.height + tableView!.contentInset.top - (tableView!.bounds.size.height - navigationView.bounds.size.height)
+        navigationView.shouldShowShadow = tableView!.contentOffset.y < maxY
+    }
+
+    
+    // MARK: KeyboardNotification delegate
+    
+    func keyboardNotification(notification: NSNotification) {
+        
+        if let userInfo = notification.userInfo {
```
syoung-smallwisdom: guard

BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -378,9 +378,30 @@ open class SBABaseScheduledActivityManager: NSObject, ORKTaskViewControllerDeleg
...
+        if step.stepViewControllerClass() == ORKFormStepViewController.self, let task = taskViewController.task {
+            if SBAGenericStepViewController.doesSupport(step) {
+                return instantiateGenericStepViewController(for: step, task: task, result: taskViewController.result)
+            }
         }
         
+        // If the default view controller for this step is an `ORKQuestionStepViewController` (and not a subclass)
+        // then replace that implementation with the one from this framework.
+        if step.stepViewControllerClass() == ORKQuestionStepViewController.self, let task = taskViewController.task {
+            if SBAGenericStepViewController.doesSupport(step) {
```
syoung-smallwisdom: You could just have the` instantiateGenericStepViewController()` method use a guard and check for support.

Otherwise, please remove the nested if statement.

BridgeAppSDK/SBASignUpViewController.swift
```diff
@@ -211,6 +211,38 @@ open class SBASignUpViewController : UIViewController, SBASharedInfoController,
     
     // MARK: ORKTaskViewControllerDelegate
     
+    
+    open func taskViewController(_ taskViewController: ORKTaskViewController, viewControllerFor step: ORKStep) -> ORKStepViewController? {
+        
+        // If the default view controller for this step is an `ORKInstructionStepViewController` (and not a subclass)
+        // then replace that implementation with the one from this framework.
+        if step.stepViewControllerClass() == ORKInstructionStepViewController.self, let task = taskViewController.task {
+            if SBAGenericStepViewController.doesSupport(step) {
```
syoung-smallwisdom: nested if statement

BridgeAppSDK/SBAStepChoiceCell.swift
```diff
@@ -0,0 +1,327 @@
...
+        setNeedsUpdateConstraints()
+    }
+    
+    override open func updateConstraints() {
+        
+        NSLayoutConstraint.deactivate(self.constraints)
+        
+        /*
+        // if we have a defined textField width, we use that and center the text field and ruleView horizontally.
+        // Otherwise, we pin left and right edges to the superview with some side margin
```
syoung-smallwisdom: Why is this included but commented out?

syoung-smallwisdom (changes requested):
Few more things to clean up. Almost there!

BridgeAppSDK/SBAScheduledActivityManager.swift
joshbruhin: I've refactored the `doesSupport()` method on `SBAGenericStepViewController` to do this check on the default VC class for the step.

joshbruhin: @syoung-smallwisdom ping

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,579 @@
...
+        }
+    }
+    
+    /**
+     Retrieve the current ORKStepResult.
+     @return    An ORKStepResult object with the current answers for all of its ORKFormItems
+     */
+    open func results(parentResult: ORKStepResult) -> ORKStepResult {
+        
+        if let formItems = formItemsWithAnswerFormat() {
```
syoung-smallwisdom: `guard let formItems = formItemsWithAnswerFormat() else { return self.parentResult }`

BridgeAppSDK/SBAGenericStepDataSource.swift
```diff
@@ -0,0 +1,579 @@
...
+                    let numericQuestionFormat = result as? ORKNumericQuestionResult,
+                    numericQuestionFormat.unit == nil {
+                    
+                    numericQuestionFormat.unit = numericAnswerFormat.unit
+                }
+                
+                result.startDate = answerDate
+                result.endDate = answerDate
+                
+                qResults.append(result)
```
syoung-smallwisdom: Just use `parentResult.addResult(result)` here.

BridgeAppSDK/SBAStepHeaderView.swift
syoung-smallwisdom: @joshbruhin ping

syoung-smallwisdom (changes requested):
So very close. Please ping again when questions answered. Thanks!

joshbruhin:
@syoung-smallwisdom changes made. Also had to make a change to `updateAnswers()` method in `SBAGenericStepDataSource` because in Elevate, I was passing in an ORKStepResult. But in Bridge, we're passing in an ORKTaskResult. So that method now find the relevant ORKStepResult

syoung-smallwisdom:
Tests are failing. Can you please revert Registration step  to old design and use a factory override at the app level. We can look at generalized version later. Thanks 

Also, looks like there are a few places where I have a question that still hasn't been answered. I can take a look tonight but it would be nice to pull this in today.

joshbruhin:
@syoung-smallwisdom not sure I understand. If we revert registration step to old design - in other words, don't use generic UI - what do we need a factory override at the app level for?

syoung-smallwisdom:
The failed unit tests are in the registration. Older apps commonly include profile info that is pulled from HealthKit and consequently, not supported by the generic step. I'm guessing that is why it's failing. What I propose is reverting SBARegistrationStep to the older methods here. Then in elevateMS, leave in the code that currently works to show the generic step view controller instead. We can look into generalization later when HealthKit is supported at this level.

syoung-smallwisdom:
Basically, I'm looking for what to cut from this PR in the interest of trying to get this in today. Otherwise, we will want to hold off on pulling this in until after release.

screen shot 2017-07-24 at 5 59 24 pm

joshbruhin:
@syoung-smallwisdom changing parent of 'SBARegistrationStepVC' back to 'ORKFormStepVC' seemed to resolve the CI failure. Looking now at the remaining open questions

syoung-smallwisdom:
@joshbruhin Thanks for the update.

BridgeAppSDK/SBAStepHeaderView.swift
joshbruhin: @syoung-smallwisdom trying to figure out how to set preferredMaxLayoutWidth to deal with screen rotation. Happen to know? I notice now that the app is portrait only, but in my testing earlier when the app was rotating, the labels did properly resize, including the height. I'll keep researching...

BridgeAppSDK/UIView+Layout.swift
joshbruhin: fixed

BridgeAppSDK/SBAVisualConsentStep.swift
joshbruhin: @syoung-smallwisdom this is done

BridgeAppSDK/SBAStepHeaderView.swift
syoung-smallwisdom: Hmmm... not sure. It's not a huge deal for our current apps (which are constrained to portrait), but something to keep in mind for iPad, etc.

syoung-smallwisdom (approved):

syoung-smallwisdom:
NOTE: This PR has been approved but is on hold until we discuss whether or not we want to include it prior to releasing v1.0 of JourneyPRO and elevateMS.

syoung-smallwisdom:
Changes look good. Thanks for updating. Will pull this in *after* build 23 goes out (tonight or tomorrow am)

joshbruhin:
@syoung-smallwisdom I submitted a minor post-merge fix to Elevate. Everyone will need that if they update BridgeAppSDK. Other than that, I couldn't find any other issues. I do have some issues in the Elevate "custom UI" branch. I'll ping you when all that's fixed and ready to merge.

PR #281, opened 2017-07-12
logout_notifications

dephillipsmichael:
Remove local notifications when user data is reset

### DIFF
--- a/BridgeAppSDK/SBAUser.swift
+++ b/BridgeAppSDK/SBAUser.swift
@@ -34,6 +34,7 @@
 import UIKit
 import BridgeSDK
 import ResearchKit
+import UserNotifications
 
 /**
  The `SBAUser` model object is intended as a singleton object for storing information about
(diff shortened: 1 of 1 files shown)


BridgeAppSDK/SBAUser.swift
```diff
@@ -53,6 +54,7 @@ public final class SBAUser: NSObject, SBAUserWrapper, SBANameDataSource, SBBAuth
         lockQueue.async {
             self.resetUserDefaults()
             self.resetKeychain()
+            self.resetLocalNotifications()
```
syoung-smallwisdom: Doesn't need to be in the lock queue.

syoung-smallwisdom (approved):
Can you move the call outside the lock queue? Otherwise, looks good.

dephillipsmichael:
@syoung-smallwisdom moved outside the lock

PR #286, opened 2017-07-18
Update pointer and move survey creation to background thread

syoung-smallwisdom:
There are a number of crashes reported in iTunes Connect when the survey factory is trying to create the survey, but the crash logs do not return an exception for *why* the app is crashing, just that it is doing so on the main thread. Since we’ve received reports that the daily check-in is crashing the app, and that survey has a lot of customization, I am suspicious that the app is crashing b/c it was blocking the main thread (taking too long - it looks at data groups which is an expensive operation) so I’ve moved building the survey to the background queue.  Additionally, as a pair of suspenders, I am explicitly retaining the references. Swift should do this automagically, but figured it wouldn’t hurt. :)

### DIFF
--- a/BridgeAppSDK/SBASurveyFactory.swift
+++ b/BridgeAppSDK/SBASurveyFactory.swift
@@ -68,6 +68,12 @@ open class SBASurveyFactory : SBABaseSurveyFactory {
         return json["steps"] as? [NSDictionary]
     }()
     
+    // This is a time-consuming operation (b/c data groups have moved to the keychain)
+    // so use a lazy load to only load when called, but then retain the result in memory.
+    lazy open var currentDataGroups: Set<String> = {
+        return Set((UIApplication.shared.delegate as? SBAAppDelegate)?.currentUser.dataGroups ?? [])
+    }()
+    
     /**
      Factory method for creating an ORKTask from an SBBSurvey
      @param survey      An `SBBSurvey` bridge model object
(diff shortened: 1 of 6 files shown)


BridgeAppSDK/SBASurveyFactory.swift
```diff
@@ -68,6 +68,12 @@ open class SBASurveyFactory : SBABaseSurveyFactory {
         return json["steps"] as? [NSDictionary]
     }()
     
+    // This is a time-consuming operation (b/c data groups have moved to the keychain)
+    // so use a lazy load to only load when called, but then retain the result in memory.
+    lazy open var currentDataGroups: Set<String> = {
```
Erin-Mounts: The most current data groups from Bridge can be directly accessed on the SBBStudyParticipant record through the "participantDataGroups" profile item, as a Set\<String\>. Access will be fast, as the study participant record is retrieved from CoreData and cached in memory when setting up Bridge initially, and then kept up-to-date with any changes.

Erin-Mounts (approved):
one comment about an alternative way to access data groups, otherwise looks good, merge when ready

BridgeAppSDK/SBASurveyFactory.swift
syoung-smallwisdom: How do I access the SBBStudyParticipant singleton?

Erin-Mounts: You shouldn't need to access it directly? You can access its properties through the profile manager, and they're all set up already as ProfileItems in either BridgeAppSDK (for the base properties) or the app (for the custom attributes). The one for data groups has the profile key "participantDataGroups".

PR #288, opened 2017-07-19
Fix funkiness at end of BBL tasks

Erin-Mounts:
- make sure to remove BBL view/view controller from hierarchy
- force the phone back to portrait orientation before moving on to the completion step


syoung-smallwisdom (approved):
looks interesting. ;)

PR #292, opened 2017-07-24
mood_survey_bug

dephillipsmichael:
Fix bug where pre-populating mood survey was one off

### DIFF
--- a/BridgeAppSDK/SBAMoodScaleStepViewController.swift
+++ b/BridgeAppSDK/SBAMoodScaleStepViewController.swift
@@ -200,7 +200,7 @@ open class SBAMoodScaleStepViewController: ORKStepViewController {
         else {
             return
         }
-        _selectedIndex = moodScale.intValue - 1
+        _selectedIndex = moodScale.intValue
     }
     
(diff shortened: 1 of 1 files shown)


syoung-smallwisdom (changes requested):
I believe that the bug is setting the value at 0. Please verify what scale is used by ResearchKit and look at updating the server-based survey. Thanks.

dephillipsmichael:
@syoung-smallwisdom can you elaborate? I'm not really following that the bug is setting the value at 0.  The bug is easily reproducible by doing on question on the daily check in, and then pressing back.  You will see the answer on that screen is one off.

syoung-smallwisdom:
Can you please assign the ticket to me and I will take a look. Thanks!

syoung-smallwisdom:
Thanks for bringing this to my attention. Unfortunately, the fix for this requires allowing the mood scale questions to be a different result type so that the server will get the correct result. Will require additional work at Journey PRO level to show in the dashboard. Sorry.

See https://github.com/Sage-Bionetworks/BridgeAppSDK/pull/293

Closing this PR.

PR #293, opened 2017-07-24
Fix the mood scale questions to return result in expected format.

syoung-smallwisdom:
Requires: https://github.com/Sage-Bionetworks/ResearchUXFactory-iOS/pull/67


Erin-Mounts (approved):
looks good

PR #295, opened 2017-07-24
If the user signs consent but sharing is “none” then change to “study”

syoung-smallwisdom:
Journey PRO has the sharing scope of "none". Should be "study" once the user has signed consent.


Erin-Mounts (approved):
looks good

PR #296, opened 2017-07-24
IA 412 - animation time

EricSiegNW:
change default animation time from 2 seconds to 1 second

This is the latest issue that I remember it being mentioned, but there have been random complaints about animations being slow. Moving the default time from 2 to 1 makes it feel a lot snappier and would seem to resolve the complaints.


syoung-smallwisdom (approved):
oops, right.

Actually, isn't standard crossfade 0.3?

PR #308, opened 2017-07-28
public_reloading

dephillipsmichael:
This is causing a critical bug with elevateMS where loadFromServerAll is not happening.  It's not necessarily an issue with SBAScheduledActivityManager, but how I am sub-classing it

Allowing reloading variable to be set by subclasses outside the framework. 

The other option is to switch up this code...

DispatchQueue.main.async(execute: {
                self?._loadedDateRange = (start, toDate)
                self?.load(scheduledActivities: scheduledActivities)
                self?.reloading = false
            })

 and move "reloading = false" to before the call to load

### DIFF
--- a/BridgeAppSDK/SBAScheduledActivityManager.swift
+++ b/BridgeAppSDK/SBAScheduledActivityManager.swift
@@ -190,7 +190,7 @@ open class SBABaseScheduledActivityManager: NSObject, ORKTaskViewControllerDeleg
             })
         }
     }
-    fileprivate var reloading: Bool = false
+    public var reloading: Bool = false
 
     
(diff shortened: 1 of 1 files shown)


syoung-smallwisdom (commented):
I'm reluctant to make this read/write. I think I would prefer overriding the reload method like we do in journeyPro. Basically, you shouldn't be mutating this array from concurrent calls. Let me think on how we can standardize on this in the superclass.

dephillipsmichael:
I can see why you are reluctant.  Let me look at overriding the reload method.

syoung-smallwisdom:
closing

PR #309, opened 2017-07-28
Fix the default button width for ResearchKit forms

syoung-smallwisdom:
There are a number of layout issues that have been introduced into Journey PRO as a result of the gigantic PR. This fixes one of them.

### DIFF
--- a/BridgeAppSDK/SBARoundedButton.swift
+++ b/BridgeAppSDK/SBARoundedButton.swift
@@ -36,7 +36,8 @@
 import UIKit
 
 public let SBARoundedButtonDefaultHeight: CGFloat = 52.0
-public let SBARoundedButtonDefaultWidth: CGFloat = 144.0
+public let SBARoundedButtonDefaultWidthWith2Buttons: CGFloat = 144.0
+public let SBARoundedButtonDefaultWidthWith1Button: CGFloat = 250.0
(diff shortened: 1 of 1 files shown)


BridgeAppSDK/SBARoundedButton.swift
```diff
@@ -79,7 +80,16 @@ public let SBARoundedButtonDefaultCornerRadius: CGFloat = SBARoundedButtonDefaul
     }
     
     public required init() {
-        super.init(frame: CGRect(x: 0, y: 0, width: 250, height: 51))
+        super.init(frame: CGRect(x: 0, y: 0, width: SBARoundedButtonDefaultWidthWith1Button, height: SBARoundedButtonDefaultHeight))
+        commonConstraintSetup(1)
+    }
+    
+    public init(numberOfButtons: Int) {
+        super.init(frame: CGRect(x: 0, y: 0, width: SBARoundedButtonDefaultWidthWith2Buttons, height: SBARoundedButtonDefaultHeight))
```
Erin-Mounts, line 88: should this check numberOfButtons to choose what value to set for width in super.init(frame:)?

Erin-Mounts (approved):
one question, otherwise looks good, merge when ready

BridgeAppSDK/SBARoundedButton.swift
syoung-smallwisdom: no, this is a work-around to Josh's stuff and will be updated using constraints.

PR #310, opened 2017-07-28
Load from cache first and then load from server

syoung-smallwisdom:
Needed by Michael.

Replaces: https://github.com/Sage-Bionetworks/BridgeAppSDK/pull/308


Erin-Mounts (approved):
looks good

PR #311, opened 2017-07-28
bug/future_loading

dephillipsmichael:
Fixed minor bug in loading the future from the server

### DIFF
--- a/BridgeAppSDK/SBAScheduledActivityManager.swift
+++ b/BridgeAppSDK/SBAScheduledActivityManager.swift
@@ -319,6 +319,16 @@ open class SBABaseScheduledActivityManager: NSObject, ORKTaskViewControllerDeleg
      @return                            The filtered list of activities
      */
     open func filteredSchedules(scheduledActivities: [SBBScheduledActivity]) -> [SBBScheduledActivity] {
+        if _loadingState == .fromServerWithFutureOnly {
+            // The future only will be in a state where we already have the cached data,
+            // And we will probably just be appending new data onto the cached data
+            let filteredActivities = scheduledActivities.filter({ (activity) in
+                return !self.activities.contains(where: { (existingActivity) -> Bool in
+                    return existingActivity.guid == activity.guid
+                })
+            })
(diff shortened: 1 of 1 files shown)


BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -319,6 +319,16 @@ open class SBABaseScheduledActivityManager: NSObject, ORKTaskViewControllerDeleg
      @return                            The filtered list of activities
      */
     open func filteredSchedules(scheduledActivities: [SBBScheduledActivity]) -> [SBBScheduledActivity] {
+        if _loadingState == .fromServerWithFutureOnly {
+            // The future only will be in a state where we already have the cached data,
+            // And we will probably just be appending new data onto the cached data
+            let filteredActivities = scheduledActivities.filter({ (activity) in
```
Erin-Mounts, line 325: I thought it was the other way around, it would pull future stuff first in case it *didn't* have cached data, and then go backfill the cache in the background? Or is this a different case and .fromServerWithFutureOnly isn't how that scenario is handled?

Erin-Mounts (commented):
I don't really understand why one case would need to filter differently than another, or if it does, why we don't just use a different function since it's behaving so differently...?

BridgeAppSDK/SBAScheduledActivityManager.swift
dephillipsmichael: Consider this scenario... The user has been using their app for 2 months but hasn't used the app for a week, and they enter the app.  The cache will load 2 months of data correctly at first, and then when it loads the future from the server, the 2 months of data will be deleted and replaced with just the future data.  Then, after a long time of getting 2 months of data from the server (like 20 seconds sometimes), it will have all the correct data again. Does that make sense?  I'm trying to avoid that scenario by just merging future data loaded from the server, with the existing cached data.

Erin-Mounts: ok after discussing on slack I get what's going on now

Erin-Mounts (approved):

PR #315, opened 2017-08-02
Fix Brain Baseline orientation hacks

Erin-Mounts:
so it works with other than iPhone/portrait-only as the default mode


syoung-smallwisdom (approved):
looks good

PR #317, opened 2017-08-08
Bugfix - custom view in SBAStepHeaderView

joshbruhin:
When customView is set, change to remove the existing customView from its superview, if it already exists.


joshbruhin:
@syoung-smallwisdom ping

syoung-smallwisdom (approved):
looks good

PR #320, opened 2017-08-25
rough_estimated_progress

dephillipsmichael:
Added a method to calculate the rough estimated progress of a survey task.


syoung-smallwisdom (changes requested):
Please implement as an extension in whatever app needs this functionality. The general use-case is not really appropriate.

dephillipsmichael:
This is the 3rd app I've developed that we have run into this issue, so I'd argue it is a general use-case.  But I agree it can easily be done per app with your approach so I'm going to close this.

syoung-smallwisdom:
If it should be per app, then the `ORKTask` progress func should change *or* the `ORKTask` protocol should include an optional method for estimated number of steps. :)

PR #321, opened 2017-09-01
bug/custom_continue_button

dephillipsmichael:
Allowed custom continue button to be applied to step nav view.
I also added the ability to change the progress view colors.  I thought it was going to be as simple as overriding the color extension's values but you cannot override class var for extensions like that unfortunately.

### DIFF
--- a/BridgeAppSDK/SBAStepProgressView.swift
+++ b/BridgeAppSDK/SBAStepProgressView.swift
@@ -42,6 +42,11 @@ import UIKit
 
 open class SBAStepProgressView: UIView {
     
+    // These can be used to set the global color for the progress bar
+    // If you want a different individual color per progress view, you will need to subclass
+    open static var progressColor = UIColor.headerViewProgressBar
+    open static var progressBackgroundColor = UIColor.headerViewProgressBackground
+    
     /**
      The current step in the current flow
      */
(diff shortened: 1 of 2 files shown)


BridgeAppSDK/SBAStepProgressView.swift
```diff
@@ -121,8 +126,8 @@ open class SBAStepProgressView: UIView {
         progressView.translatesAutoresizingMaskIntoConstraints = false
         stepCountLabel.translatesAutoresizingMaskIntoConstraints = false
         
-        backgroundView.backgroundColor = UIColor.headerViewProgressBackground
-        progressView.backgroundColor = UIColor.headerViewProgressBar
+        backgroundView.backgroundColor = SBAStepProgressView.progressBackgroundColor
+        progressView.backgroundColor = SBAStepProgressView.progressColor
```
syoung-smallwisdom: Subclassing and overriding will not return the progress bar background color. For that, you need to do this instead:

`type(of: self).progressBackgroundColor`

syoung-smallwisdom (changes requested):

BridgeAppSDK/SBAStepProgressView.swift
dephillipsmichael: That's a good tip, will make the change

dephillipsmichael: fixed

syoung-smallwisdom (approved):
looks good

PR #324, opened 2017-09-20
Add view with a crosshatch as per designs

syoung-smallwisdom:
Apparently the background for the colored screens is *not* suppose to be a gradient.


Erin-Mounts (approved):
Hard to visualize what it will look like, but seems fine to me

PR #326, opened 2017-09-25
feature/IA-398

dephillipsmichael:
Add functionality in base schedule manager to be notified when a task becomes available.  This can be used in conjunction with deep linking for a stable way to send users to a task or task group after they arrive in the app.

### DIFF
--- a/BridgeAppSDK/SBAScheduledActivityManager.swift
+++ b/BridgeAppSDK/SBAScheduledActivityManager.swift
@@ -92,6 +92,12 @@ open class SBABaseScheduledActivityManager: NSObject, ORKTaskViewControllerDeleg
     }
     fileprivate var _user: SBAUserWrapper!
     
+    /**
+     This closure will be non-nil when notifyWhenTaskIsAvailable is called and we
+     are waiting for a certain task to become available
+    */
+    fileprivate var notifyTaskAvailableId: String?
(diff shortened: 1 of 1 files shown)


BridgeAppSDK/SBAScheduledActivityManager.swift
```diff
@@ -1075,6 +1088,25 @@ open class SBABaseScheduledActivityManager: NSObject, ORKTaskViewControllerDeleg
         }
     }
     
+    /**
+     If the task is available right away, the callback will be invoked
+     Otherwise, we will save the callback to be executed later when today's task is available
+    */
+    public func notifyWhenTaskIsAvailable(taskId: String, callback: @escaping () -> Void) {
```
syoung-smallwisdom, line 1095: If not intended to be overridable, mark it as final.  

Additionally, please don't call this from the private load and set the callback and id again. And what happens if the developer sets this up to listen for more than one task id? Then this will get clobbered and end up with the closure not getting called at all.

syoung-smallwisdom (changes requested):
I'm not sure I understand the value of this. Why not just have the app override `open func load(scheduledActivities: [SBBScheduledActivity])` and check there?

dephillipsmichael:
Closing to be implemented within MasterScheduleActivityManager

PR #330, opened 2017-10-05
play audio when instruction step is shown

Erin-Mounts:
if there is any to play

requires https://github.com/Sage-Bionetworks/ResearchUXFactory-iOS/pull/76


syoung-smallwisdom (approved):
Looks good

PR #333, opened 2017-10-30
Refactor the activity archive to allow building an archive from a ResearchSuite result.

syoung-smallwisdom:
Move some code around and add protocols to allow using the same code to archive either an ORKTaskResult or a RSDTaskResult.


Erin-Mounts (approved):
looks good

PR #334, opened 2017-10-31
Fix access to shared user to allow calling from background thread

syoung-smallwisdom:
Change the shared user access to not use the UIApplication delegate.


Erin-Mounts (approved):
looks good

PR #335, opened 2017-12-07
Comment out assertion that fails during onboarding

syoung-smallwisdom:
This assertion was blocking me from being able to sign in to the CRF app using an existing external ID.

### DIFF
--- a/BridgeAppSDK/SBAProfileItem.swift
+++ b/BridgeAppSDK/SBAProfileItem.swift
@@ -585,7 +585,7 @@ open class SBAStudyParticipantProfileItem: SBAStudyParticipantCustomAttributesPr
     override open func storedValue(forKey key: String) -> Any? {
         guard let studyParticipant = SBAStudyParticipantProfileItem.studyParticipant
             else {
-                assertionFailure("Attempting to read \(key) (\(profileKey)) on nil SBBStudyParticipant")
+                //assertionFailure("Attempting to read \(key) (\(profileKey)) on nil SBBStudyParticipant")
                 return nil
         }
         // special-case handling for an attribute to call through to the superclass implementation


Erin-Mounts:
This is an assertion because it should never happen. Until you're signed in, the BridgeSDK will give you a dummy SBBStudyParticipant to play with. Is it maybe somehow getting called before BridgeSDK has been set up? FWIW I've successfully signed in to CRF using an existing external ID so it seems like there must be more going on here than just that.

syoung-smallwisdom:
Yup, likely there is. New phone running iOS 11.2. Honestly, didn't spend too much time on it. Let's leave the PR open for now as a marker to revisit.

syoung-smallwisdom:
@Erin-Mounts 
I had this happen again Friday. It appears to be related to trying to run the app on a device or simulator where the keychain has been blown away but there is lingering info in cache. The result is partially signed in user.

This is going to be an issue for any of our longer-running studies where the user may switch phones because the keychain isn't restored. (I had to re-login to everything when I upgraded my phone.)

Erin-Mounts:
I also saw this once on Friday, when I got most of the way through onboarding and then canceled and then tried to go through onboarding again (can’t remember if I quit/killed the app in between or not). I was able to “solve" it by deleting and re-installing but I want to figure out what exactly is getting it into this state. It should never be possible to have a SBBUserSessionInfo object with a nil studyParticipant field but that seems to be what’s happening (and at least the proximal cause of the crash).

syoung-smallwisdom:
Nice to know that I'm not the only one.  :)  And yes, deleting and reinstalling fixed the issue.

Erin-Mounts:
Welp, I was going to try to debug through this and take care of it but I can no longer reproduce it with the latest CRFValidationApp commit in /master.

Erin-Mounts:
It's getting a SBBUserSessionInfo with a nil SBBStudyParticipant field because (a) there's a non-nil SBBUserSessionInfo in CoreData cache but (b) it doesn't have the login credentials anymore, which it needs to decrypt the StudyParticipant from CoreData.

If the app logged out from Bridge, the cache would have been cleared so I'm guessing that when you "cancel" during onboarding after verifying the external ID (the point at which it logs in), it's clearing the keychain but not actually calling BridgeSDK to log you out. Possibly we assumed you wouldn't be logged in until completing onboarding because that's how apps have worked in the past.

Erin-Mounts:
P.S. this only happens if you cancel partway through onboarding (after giving the external ID twice) and then quit and restart the app; until then it has the decrypted SBBStudyParticipant in its in-memory cache and doesn't need the credentials to decrypt it.

syoung-smallwisdom:
Closing. haven't run into the issue in a while.

PR #341, opened 2018-05-08
Make 'userHasContinued' open so it can be overridden from subclass

joshbruhin:
Needed for changes on Geisinger project


syoung-smallwisdom (approved):
looks good

PR #345, opened 2019-08-13
Update xcode10.3

syoung-smallwisdom:
This will now run on Xcode 10.3

Note: BridgeSDK is still pointing at a older version that uses the auth delegate b/c I wasn't sure how, for an existing app with users who are signed in, to migrate to the newer framework version.

Requires: https://github.com/Sage-Bionetworks/ResearchUXFactory-iOS/pull/80


Erin-Mounts (approved):
👍

PR #346, opened 2019-08-19
Remove references to auth delegate and update BridgeSDK pointer

syoung-smallwisdom:
¯\\\_(ツ)\_/¯ 

I suspect that anyone using this code will have a bit of work to do that is custom to their app in order for the keychain to work.

### DIFF
--- a/BridgeAppSDK/SBBScheduledActivity+Utilities.swift
+++ b/BridgeAppSDK/SBBScheduledActivity+Utilities.swift
@@ -86,23 +86,23 @@ public extension SBBScheduledActivity {
         return self.activity.survey?.identifier
     }
     
-    /**
-     Returns either the `SBBTaskReference` or `SBBSurveyReference` identifier.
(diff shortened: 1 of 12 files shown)


BridgeAppSDK/SBBScheduledActivity+Utilities.swift
```diff
@@ -86,23 +86,23 @@ public extension SBBScheduledActivity {
...
-    @objc dynamic var scheduleIdentifier: String {
-        // Strip out the unique part of the guid
-        if let range = self.guid.range(of: ":") {
-            return String(self.guid[..<range.lowerBound])
-        }
-        else {
-            return self.guid
-        }
-    }
+//    /**
```
Erin-Mounts: commented-out code

I realize we don't _intend_ to keep maintaining this, but....

Erin-Mounts (approved):
one nit, looks good

